PR #36, opened 2021-01-29
Release/web/snowflake/1.0.0

adatzer:


adatzer:
For visibility, the previous forced update, changed this line: [link](https://github.com/snowplow/data-models/blob/85d5fbb068783a058432b4d75f69fb2fab656b77/web/v1/snowflake/sql-runner/sql/standard/02-page-views/01-main/07-commit-page-views.sql#L37)
It changes the join key for upserting to derived.page_views from `page_view_id` to `domain_sessionid`.
The reason is an edge case, where a duplicate page-view event is introduced in subsequent run. Using `domain_sessionid` ensures late-found duplicate page views are removed from the derived.page_views table.

web/v1/snowflake/sql-runner/configs/datamodeling.json
```diff
@@ -0,0 +1,58 @@
+{
+  "schema": "iglu:com.snowplowanalytics.datamodeling/config/jsonschema/1-0-0",
+  "data": {
+    "dagName": "standard-webmodel",
+    "enabled": true,
+    "storage": "Default",
+    "sqlRunner": "0.9.3",
+    "lockType": "soft",
```
bill-warner: BQ and RS are set to `"lockType": "hard"`. We should align across all the models irrespective of the value we choose.

colmsnowplow:
> It changes the join key for upserting to derived.page_views from page_view_id to domain_sessionid.
> The reason is an edge case, where a duplicate page-view event is introduced in subsequent run. Using domain_sessionid ensures late-found duplicate page views are removed from the derived.page_views table.

It sounds like we may not be addressing the root issue with this fix. I've just begun my review but please note down the details/steps to reproduce the issue in case we need to look at it more closely.

(Maybe it's fine, just need to understand it better)

.scripts/run_config.sh
```diff
@@ -22,21 +22,27 @@ do
...
 
 # Use jq to grab playbooks into a bash array in order of appearance, and database target into a var.
 playbooks=($(cat $CONFIG_PATH | jq -r '.data.playbooks[].playbook | @sh' | tr -d \'))
 target=$(cat $CONFIG_PATH | jq -r '.data.storage')
-model_path="$(dirname "${CONFIG_PATH}")/.."
+model_path=$( cd "$(dirname "$(dirname "${CONFIG_PATH}")")" && pwd -P )
 
 set -e
 
+if [ -d "${root_path}/tmp" ]; then
```
colmsnowplow: Why require that the `tmp` dir not exist, and destroy everything in it, rather than just creating it if it doesn't exist?

.scripts/pr_check.sh
```diff
@@ -16,17 +16,12 @@ done
...
-export BIGQUERY_CREDS=${BIGQUERY_CREDS:-$CREDENTIALS}
-export REDSHIFT_PASSWORD=$CREDENTIALS
-export SNOWFLAKE_PASSWORD=$CREDENTIALS
-
 echo "pr_check: Starting 10 e2e iterations"
 
 for i in {1..10}; do
   echo "pr_check: Starting e2e run $i";
 
-  bash $script_path/e2e.sh -b $SQL_RUNNER_PATH -d $DATABASE || exit 1;
```
colmsnowplow: Does this require that we pass credentials as an argument?

I think we can solve the problem by just doing what we've done for BQ:

`export REDSHIFT_PASSWORD=${DB_PASSWORD:-CREDENTIALS}`

Which means we may use either env var or pass as argument (which is what we do for the other scripts I think)

web/v1/snowflake/sql-runner/configs/datamodeling.json
colmsnowplow: Agreed. I think the thinking in that decision for the other DBs is that until they configure email alerts, it should be hard locked. When they configure email alerts they (or we) can swap it to soft locks.

Open to challenge on that plan.

web/v1/snowflake/sql-runner/sql/custom/02-page-views-join/XX-destroy/XX-destroy-page-views-join.sql
```diff
@@ -0,0 +1,37 @@
...
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+
+DROP TABLE IF EXISTS {{.output_schema}}.page_views_join{{.entropy}};
+
+-- Snowflake supports overloading of procedure names.
+-- So the data types of the arguments are needed to identify the procedure to drop.
+DROP PROCEDURE IF EXISTS {{.output_schema}}.mk_transaction(VARCHAR);
```
colmsnowplow: Not sure we should drop the procedures in the custom steps - I would keep these kinds of statements to the standard steps for simplicity I think. 

(Maybe there is a simpler/more intuitive design to what we've done in the other databases though)

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,426 @@
...
+/*
+  MK_TRANSACTION
+  Side effects procedure.
+  Input: a concatenation of one or more DML sql statements split by semicolon.
+  It is important that the statements are not DDL.
+  Either all the statements in the block succeed and get committed or all get rolled back.
+  To drop:
+  DROP PROCEDURE {{.output_schema}}.mk_transaction(VARCHAR);
+*/
+CREATE OR REPLACE PROCEDURE {{.output_schema}}.mk_transaction(DML_STATEMENTS VARCHAR)
```
colmsnowplow, line 27: What's the value of having a procedure for this instead of just wrapping the SQL in a `BEGIN .. END`  transaction?

web/v1/snowflake/sql-runner/sql/custom/02-page-views-join/XX-destroy/XX-destroy-page-views-join.sql
colmsnowplow: OK I found the docs you provide and looked into things like this: https://community.snowflake.com/s/question/0D50Z00008BDcR0SAL/how-do-you-commit-a-transaction-if-there-are-no-problems-but-roll-it-back-if-there-are

Makes sense now.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
colmsnowplow: OK I found the docs you provide and looked into things like this: https://community.snowflake.com/s/question/0D50Z00008BDcR0SAL/how-do-you-commit-a-transaction-if-there-are-no-problems-but-roll-it-back-if-there-are

Makes sense now.

web/v1/snowflake/sql-runner/configs/datamodeling.json
adatzer: I totally agree on the consistency argument. My reasoning here is that since the model is safe to rerun, maybe it is worth denoting that too, or at least add this to the documentation.

.scripts/run_config.sh
adatzer: Strictly speaking there is no need. But assuming the dir structure of the repo, and considering `/tmp` to be temporarily there for credentials, i would prefer it to have separate concerns that e.g. output or any other pre-existing dir. Of course i could change that back, if we think there is no real reason for this distinction.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,426 @@
...
+  SRC_SCHEMA:  the schema of the source table
+  SRC_TBL:     the source table name
+  TRG_SCHEMA:  the schema of the target table
+  TRG_TBL:     the target table name
+  AUTOMIGRATE: whether any extra columns in source table will be added in target table (only 'TRUE' enables)
+
+  To drop:
+  DROP PROCEDURE {{.output_schema}}.column_check(VARCHAR,VARCHAR,VARCHAR,VARCHAR,VARCHAR);
+*/
+CREATE OR REPLACE PROCEDURE {{.output_schema}}.column_check(SRC_SCHEMA  VARCHAR,
```
colmsnowplow: This is good stuff. I've found a slight wrinkle - it needs to make sure that new columns are in the right position (at the end of the table), and that existing columns all exist _in the right order_ - otherwise our insert will fail:

```
CREATE TABLE scratch.test1 AS(SELECT 1 AS one, 2 AS two);

CREATE TABLE scratch.test2 AS(SELECT 1 AS one, 3 AS three);

CREATE TABLE scratch.test3 AS(SELECT 1 AS one, 1.5 AS onepointfive, 2 AS two)

-- This will fail, as we expect it to:
CALL derived.column_check(
    UPPER('scratch'),
    UPPER('test2'),
    UPPER('scratch'),
    UPPER('test1'),
    'TRUE'
);

-- This will succeed, and will add 'onepointfive' column to the end of the table
CALL derived.column_check(
    UPPER('scratch'),
    UPPER('test3'),
    UPPER('scratch'),
    UPPER('test1'),
    'TRUE'
);


SELECT * FROM scratch.test1
-- result attached
```
screenshot 2021-02-01 at 18 11 51

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,426 @@
...
+          return "ok. Columns match."
+      }
+  }
+
+  // == Helpers ==
+
+  function list_cols_with_type(sch,tbl,delimiter) {
+      var stmt = `
+          SELECT
+              LISTAGG(
```
colmsnowplow, line 124: It might make sense to return an array, and LISTAGG later. I'm not tied to it though, whatever solution works best.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,426 @@
...
+  DRV_UPSERT_LOOKBACK FLOAT: corresponds to {{.upsert_lookback_days}}::float
+
+  To drop:
+  DROP PROCEDURE {{.output_schema}}.commit_with_metadata(VARCHAR,VARCHAR,VARCHAR,VARCHAR,
+                                                         VARCHAR,VARCHAR,VARCHAR,VARCHAR,
+                                                         VARCHAR,VARCHAR,VARCHAR,VARCHAR,
+                                                         VARCHAR,VARCHAR,VARCHAR,FLOAT);
+
+*/
+CREATE OR REPLACE PROCEDURE {{.output_schema}}.commit_with_metadata(
```
colmsnowplow: I'm not sure I'm a fan of tying the commit to updating the metadata table - because for custom modules, I'm not sure we should force the customer to care about the metadata tables. 

Open to discussion.

.scripts/pr_check.sh
adatzer: The `run_config.sh`, which is called by `e2e` is already handling the credentials in this way (allow both as arg and as env vars (https://github.com/snowplow/data-models/blob/85d5fbb068783a058432b4d75f69fb2fab656b77/.scripts/run_config.sh#L48)), so i am not sure if we need to repeat here. Certainly i will look at it again.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,426 @@
...
+                                                        STG_TRG_TBL         VARCHAR,
+                                                        STG_JOIN_KEY        VARCHAR,
+                                                        SKIP_DRV            VARCHAR,
+                                                        DRV_SRC_SCHEMA      VARCHAR,
+                                                        DRV_SRC_TBL         VARCHAR,
+                                                        DRV_TRG_SCHEMA      VARCHAR,
+                                                        DRV_TRG_TBL         VARCHAR,
+                                                        DRV_JOIN_KEY        VARCHAR,
+                                                        DRV_PARTITION_KEY   VARCHAR,
+                                                        DRV_UPSERT_LOOKBACK FLOAT)
```
colmsnowplow: I think this would be a lot more readable if we used full names for the input variables.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,426 @@
...
+  COLUMN_CHECK
+  Checks for mismatched columns between source and target tables.
+  If source table is missing columns, it errors.
+  If AUTOMIGRATE is 'TRUE',
+    it allows the target table to get added the additional columns,if any, in source.
+  Since ALTER TABLE is a DDL statement, it will be its own transaction.
+  This means that even if it was placed inside another transaction,
+    it would commit that, and then start another(implicit) for its own execution.
+  So, also it cannot be explicitly rolled back.
+  This is why column_check is not part of the commit_with_metadata procedure.
```
colmsnowplow: Could commit_table just contain two distinct transactions - one to check and add columns, and another to update the table?

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,426 @@
...
+      var drv_src_columns = list_cols(DRV_SRC_SCHEMA,DRV_SRC_TBL);
+      if (cols_check(drv_trg_columns, drv_src_columns) === false) {
+          throw Error("ERROR: column mismatch:" + DRV_TRG_TBL + "," + DRV_SRC_TBL);
+      }
+
+      var drv_trg_ref = `` + DRV_TRG_SCHEMA + `.` + DRV_TRG_TBL;
+      var drv_src_ref = `` + DRV_SRC_SCHEMA + `.` + DRV_SRC_TBL;
+      var drv_join_condition = `` + DRV_JOIN_KEY + ` IN
+                               (SELECT ` + DRV_JOIN_KEY + ` FROM ` + drv_src_ref + `)`;
+      var drv_lower_limit = `
```
colmsnowplow: Can we execute the query and store the value in a var?

If so it'll be more efficient to just plug the value into the query than to have the queries nested.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,426 @@
...
+      return colsArr1.every(function(c) {return colsArr2.includes(c);}) &&
+             colsArr1.length === colsArr2.length;
+  }
+
+  $$
+;
+
+
+/*
+  CTAS_SIMILAR
```
colmsnowplow: I like what you've done here - it's a clever solution. I wonder if we should consider trying to simplify things though - since we only need it once, do we need to create a procedure for this?

This is open for discussion, I'd like to think about it a bit, but one option is to just accept that base_staged has both `page_view_id` and `contexts_com_snowplowanalytics_snowplow_web_page_1`:

```
SELECT 
    contexts_com_snowplowanalytics_snowplow_web_page_1[0]:id AS page_view_id,
    *
FROM ...
```

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,426 @@
...
+  This procedure creates the target table as "similar" to the source table.
+   - Imitates BigQuery's EXCEPT(col1,...) functionality.
+   - Accepts a string of comma separated column tranformations to add
+   - and another with their respective aliases.
+
+  Input:
+  SRC_SCHEMA:    the schema of the source table
+  SRC_TBL:       the source table name
+  XCLUDE_COLS:   a string of comma-separated columns to exclude
+  TRANS_COLS:    a string of comma-separated column tranformations
```
colmsnowplow: I think it's preferable to have the transformations be coupled with the aliases if possible, if this argument would take a list of tuples, for example.

.scripts/run_config.sh
colmsnowplow: I think there are two questions: 

1. Should the script fail if `tmp` exists already? 

I don't see a good reason. Even if we expect it to not exist, it's easy to make the workflow smoother by using `mkdir -p`, which won't fail if it exists.

2. Should we drop the entire `tmp` directory instead of just deleting one file?

I can see an argument that it's safe to drop the entire thing. Originally, I wanted it to also work if someone has manually added a JSON file containing their BQ credentials - in fact one needs to do this if using great_expectations or sql-runner directly, without our scripts for example.

However I'm not sure there's a strong argument for that now, as I changed a lot since then... Let's return to it.

.scripts/pr_check.sh
colmsnowplow: So we do need to set the `SNOWFLAKE_PASSWORD` and `REDSHIFT_PASSWORD` variables in our script - otherwise great_expectations will fail if we're using only env vars. (which is the original bug)

These are distinct from `DB_PASSWORD` currently, but in hindsight there's surely a better design choice there.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
adatzer: Since `test3` is the source table, `test1` is the target, so any extra columns of `test3`, get added at the end of `test1`, with `test1`'s columns in place. Isn't it the expected output?

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
adatzer: This is indeed something i have also been considering. The idea is that if someone wants metadata, it should be simple to accommodate in a custom module.
Would it make sense to add a flag as an argument, so that at least it is there for those who would prefer to do so?

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
adatzer: If i understand the question correctly, this is what essentially happens. The `ALTER` statement executes in its own transaction and commit does the updates. So the question is whether we want to wrap both inside a stored procedure. I chose not to, so as: a) to not create the illusion of a unit and b) `column_check` as a procedure of its own for reusability independent of committing. What do you think?

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
adatzer: Yes, this should be possible. But why would that make it more efficient?

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
colmsnowplow: I don't think so - because what happens when we then insert the data?

In this case, inserting `SELECT * FROM test3` into `test1` will fail, because it will attempt to put a decimal into an integer field. 

But even if the column types are compatible - you'll end up with the data in the wrong columns, without it being obvious why.

So I think it's best to enforce a rule that you can't add columns to the middle of an existing table, only the end. If you try to, the columnCheck should fail.

adatzer: I may be wrong or misunderstanding, but actually running this, creates the same column type for `onepointfive` in `test1` as it is in `test3`. So it seems that column types are compatible.
Also, adding columns, adds at the end(I haven't found an explicit guarantee on this on the Snowflake docs, but it seems to be the case). The `column_check` code doing this is:
```
function add_columns_to(sch, tbl, cols) {
      var alter_stmt = `ALTER TABLE ` + sch + `.` + tbl + ` ADD COLUMN ` + cols;
      snowflake.createStatement({sqlText: alter_stmt}).execute();

      return "ok. Columns added.";
  }
```
So, how could we enforce what `ALTER` does? 
Then, indeed a `select *` for insert will fail, but this is something that comes after the column check/migrate. And it would also fail even even if we could enforce the column order. That's why the `commit_with_metadata` uses an explicit select matching the target columns instead of a `select *`.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
colmsnowplow: > The idea is that if someone wants metadata, it should be simple to accommodate in a custom module.

This I am on board with.

I think it's preferable to have an 'update_metadata' procedure as distinct from the commit procedure.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
colmsnowplow: Do you have an example use case for needing to use `column_check` independently of the commit table procedure?

Would help me check my assumptions. :)

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
colmsnowplow: Nested queries are inefficient in general in databases - because it has to create the query plan, execute the nested statement, and hold the output in memory for the remainder of the query (and also lookup that memory when executing the WHERE clause in this case).

In this case, it makes very little difference, because it's just one value. But it's still the more efficient structure to use.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
adatzer: Do you mean updating the metadata tables to go in its own transaction independent of the commit?
There is also the option to overload a commit procedure name, with different arguments, one with and one without committing metadata.

web/v1/snowflake/sql-runner/playbooks/custom/02-page-views-join/01-page-views-join.yml.tmpl
```diff
@@ -0,0 +1,33 @@
+:targets:
+- :name:
+  :type:      snowflake
+  :account:
+  :database:
+  :warehouse:
+  :username:
+  :password:
+:variables:
+  :scratch_schema:     scratch
```
colmsnowplow, line 10: I think the defaults should still have `_prod1` - we should follow the default that tech ops deploy for new customers. Double check with them if you're unsure.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
adatzer: There isn't one here yet. The point of it being out of commit is that it cannot be done inside an explicit transaction, since the implicit one (enforced by `ALTER`, as by any DDL statement) would break the atomicity. (not sure if this answers)

web/v1/snowflake/sql-runner/playbooks/custom/02-page-views-join/01-page-views-join.yml.tmpl
adatzer: I do not doubt this may be the case. I had initially changed them so, and that is the reason i have this in its own commit, so that it can be dropped anytime.
The reason for reverting back started from the `derived` and `scratch` references in the tests' validation configs. I assumed (maybe wrongly) that it sounds more general to have `scratch` instead of e.g. `scratch_prod1`, since we only fill those variables to signify them being required, just like the `start_date`. 
Of course, this can be changed, i guess i'm not sure whether it being `scratch` or `scratch_prod1` creates a problem.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
adatzer: This was done this way in order to avoid having to map sql types to javascript ones, with varchar being the simplest. But i can certainly try.

web/v1/snowflake/sql-runner/sql/custom/02-page-views-join/XX-destroy/XX-destroy-page-views-join.sql
adatzer: Does this apply to their "definition" too?
It is a bit redundant to add / drop in every module. This could be solved by the setup module, or by letting all those defs/drops to be handled only by base. What do you think?

web/v1/snowflake/sql-runner/sql/custom/02-page-views-join/01-main/00-setup-page-views-join.sql
```diff
@@ -0,0 +1,27 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+
+CREATE TABLE IF NOT EXISTS {{.output_schema}}.page_views_join{{.entropy}} (
+
+    page_view_id                 VARCHAR(36)      NOT NULL,
```
colmsnowplow, line 20: Nit: The other models use a single tab to indent

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/00-setup-metadata.sql
```diff
@@ -0,0 +1,46 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+
+-- Permanent metadata table
+CREATE TABLE IF NOT EXISTS {{.output_schema}}.datamodel_metadata{{.entropy}} (
+    run_id                     TIMESTAMP_NTZ,
```
colmsnowplow: Nit: Some files use single tab indent, some tabs use double. The other models use single

web/v1/snowflake/sql-runner/sql/standard/01-base/01-main/08-add-page-view-id.sql
```diff
@@ -0,0 +1,34 @@
...
+
+/*
+  Does same as what BigQuery does with:
+
+  SELECT
+    a.contexts_com_snowplowanalytics_snowplow_web_page_1_0_0[SAFE_OFFSET(0)].id AS page_view_id,
+    a.* EXCEPT(contexts_com_snowplowanalytics_snowplow_web_page_1_0_0)
+  FROM a
+*/
+CALL {{.output_schema}}.ctas_similar(
```
colmsnowplow: I think we should indeed consider simplifying. Let's discuss it once I have all the comments noted.

web/v1/snowflake/sql-runner/sql/standard/04-users/01-main/02-users-sessions-this-run.sql
```diff
@@ -0,0 +1,31 @@
...
+    SELECT
+        a.*
+
+    FROM
+        {{.output_schema}}.sessions{{.entropy}} AS a
+
+    INNER JOIN {{.scratch_schema}}.users_userids_this_run{{.entropy}} AS b
+        ON a.domain_userid = b.domain_userid
+
+    WHERE a.start_tstamp >= (SELECT MIN(start_tstamp) FROM {{.scratch_schema}}.users_userids_this_run{{.entropy}})
```
colmsnowplow: Why use nested query in this step instead of limits table?

web/v1/snowflake/sql-runner/sql/custom/02-page-views-join/XX-destroy/XX-destroy-page-views-join.sql
colmsnowplow: How we've done it in the BQ model is that the functions are defined once in the setup step, but that step is the first step of each module's playbook.

The main reason for this is that we can imagine people running the models on independent schedules, with independent configurations. 

The other models don't pass upsert_lookback as arguments to the function, but define it when the function is defined. I think that's actually design flaw.

It also doesn't drop the function as part of normal cleanup, which is a design decision we should consider. Let's add it to the list of things to figure out when we discuss.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
adatzer: I am not an expert on the matter, but it is not a correlated subquery, so it will be calculated only once. The alternative of storing to a variable means we execute an extra query calling the JavaScript API one more time. For both cases there is a constant penalty, so it could be that the total time difference would be small. My assumption here on the memory lookups is that the subquery result will be cached. Of course i am open to it and willing to change it, just testing my assumptions.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
colmsnowplow: Yes, ALTER TABLE ADD COLUMN will add to the end of the column list - you can't mutate an existing table to add columns in the middle. 

So the design pattern I suggest is that we only allow one to add a column to the end of your table, and if you try to do something different, then that should fail. If someone's doing this, I expect that they have either made a mistake (in which case it should fail), or they need the specific ordering of columns that they've provided (in which case they need to manage it manually). 

If we are reordering the columns for them, the behaviour is counterintuitive for both cases IMO. 

I think it also impairs readability of the code - if I'm tracing back from a production table that has been amended many times, I don't expect to see a different order of columns.

Those are the main considerations I have here. We can add it to the list to chat about.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
adatzer: > since we only need it once, do we need to create a procedure for this?

Indeed, we do not need. But i actually thought that this may be the most useful procedure of all, maybe not to us or yet, but it makes it easy to slightly change a table and keep/drop/transform some columns. I can imagine a use case, but sure, no apparent need per se.

> one option is to just accept that base_staged has both page_view_id and contexts_com_snowplowanalytics_snowplow_web_page_1

This is indeed an option. But if we can do it, i wonder why not.

web/v1/snowflake/sql-runner/sql/standard/04-users/01-main/02-users-sessions-this-run.sql
adatzer: From my tests, this was slightly faster. No other reason.

web/v1/snowflake/sql-runner/playbooks/custom/02-page-views-join/01-page-views-join.yml.tmpl
colmsnowplow: Neither create a problem, it makes no difference when considering this project as a whole - but the default that we deploy have `_prod1` appended to the end, so using that in the playbooks reduces the amount of changes we need to make to deploy the model for customers. 

Really it's not a big deal, it's just nice if we have consistency across all of our defaults.

adatzer:
_Address review changes I_
This commit includes:
1. Committing to metadata gets decoupled from committing to tables.
2. The custom module does not define/drop the procedures.
3. The `column_check` got included in the `commit_table` procedure (not in the transaction), instead of being a separate statement, so as to indeed reduce the duplication of checks. It was also changed to enforce the order of columns allowed to be migrated. There was no change to the arguments data types (to couple with aliases) for simplicity, since this is now only used internally. 
3. The `base_events_this_run_tmp` table was indeed not needed, so was removed. Instead, the `ctas_similar` procedure was modified, so that it allows to have the same `events_staged` columns across web models, providing a functionality that BigQuery has and Redshift doesn't need because of the federated tables.
4. Argument names made more readable and arguments' number was reduced where possible. 

(Executing the lower-limit statement separately inside commit_tables has the nuance of returning a timestamp data-type, that would map to the snowflake-internal sfDate. That is why this change did not happen.)

.scripts/run_config.sh
adatzer: @colmsnowplow i'm sorry, indeed i had misunderstood the issue. In the "Address review changes II" commit, i also tried to remove the use of `DB_PASSWORD` so as to have same env vars for both sql-runner and great-expectations.
Finally, i wasn't aware of the use case for storing in `bq_creds.json`, so, even though still a bit tempting, reverted the total removal of `tmp`. As far as i've tested the issue seems solved. Please, let me know if you also find so.

colmsnowplow: Actually I think this one has been confusing because I realised recently that we have _two_ bugs:

1. The credentials weren't handled as expected
2. The `tmp` folder wasn't created when it didn't exist.

So it's my bad for first introducing the bugs, and second missing one of them and making it more confusing.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
colmsnowplow: OK seems like a good reason to keep it as it is :)

web/v1/snowflake/sql-runner/sql/standard/04-users/01-main/02-users-sessions-this-run.sql
colmsnowplow: Ah this makes sense. So one thing that I wonder is whether we should still have the limits tables, to accommodate custom steps.

The thing about the users step is that we _don't_ stage data for this level of aggregation ahead of time, because it would be too much data. So I'm not sure on what we should do here - if people are likely to want to query the sessions table in order to aggregate to the users level, we should provide them with limits to use. 

Does this make sense?

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -100,11 +98,17 @@ CREATE OR REPLACE PROCEDURE {{.output_schema}}.column_check(SRC_SCHEMA  VARCHAR,
 
       var columnAdditions = sourceColumns.filter(notIncludedIn(targetColumns));
 
-      if ( columnAdditions.length !== 0 ) {
+      if (sourceColumns.some(notIncludedIn(targetColumns)) === true) {
           if ( AUTOMIGRATE !== 'TRUE' ) {
               throw "ERROR: Target table is missing column(s),but automigrate is not enabled.";
           } else {
-              add_columns_to(TRG_SCHEMA, TRG_TBL, columnAdditions.join(','));
+              // enforce order
```
colmsnowplow, line 105: Nice

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -156,139 +160,111 @@ CREATE OR REPLACE PROCEDURE {{.output_schema}}.column_check(SRC_SCHEMA  VARCHAR,
...
-                                                        STG_JOIN_KEY        VARCHAR,
-                                                        SKIP_DRV            VARCHAR,
-                                                        DRV_SRC_SCHEMA      VARCHAR,
-                                                        DRV_SRC_TBL         VARCHAR,
-                                                        DRV_TRG_SCHEMA      VARCHAR,
-                                                        DRV_TRG_TBL         VARCHAR,
-                                                        DRV_JOIN_KEY        VARCHAR,
-                                                        DRV_PARTITION_KEY   VARCHAR,
-                                                        DRV_UPSERT_LOOKBACK FLOAT)
+CREATE OR REPLACE PROCEDURE {{.output_schema}}.commit_tables(
```
colmsnowplow, line 183: Nit: we're only committing one table at a time - so I would name it `commit_table`

web/v1/snowflake/sql-runner/sql/standard/04-users/01-main/02-users-sessions-this-run.sql
adatzer: Yes, it certainly makes sense. Maybe we could also then add that table as a temporary output of the users module in the readme's too, so that is clear that it should exist and it can potentially be used?

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
colmsnowplow: As we discussed, doing both in one might have a negative impact on customer experience - because it's less intuitive. 

However Snowflake say that it's more performant to use one transaction instead of two if possible.

So, we'll test if there's a performance impcat to separating the two, and if not we can break this down to be more atomic and simpler. If so, it can stay as is.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -156,139 +160,111 @@ CREATE OR REPLACE PROCEDURE {{.output_schema}}.column_check(SRC_SCHEMA  VARCHAR,
...
+                                                 STAGE_NEXT            VARCHAR,
+                                                 AUTOMIGRATE           VARCHAR,
+                                                 STAGING_SOURCE        VARCHAR,
+                                                 STAGING_TARGET        VARCHAR,
+                                                 STAGING_JOIN_KEY      VARCHAR,
+                                                 SKIP_DERIVED          VARCHAR,
+                                                 DERIVED_SOURCE        VARCHAR,
+                                                 DERIVED_TARGET        VARCHAR,
+                                                 DERIVED_JOIN_KEY      VARCHAR,
```
colmsnowplow: Join key and partition key should always be the same across all 3 tables involved. So each only need to be one input.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -334,42 +300,41 @@ CREATE OR REPLACE PROCEDURE {{.output_schema}}.commit_with_metadata(
...
 
 */
-CREATE OR REPLACE PROCEDURE {{.output_schema}}.ctas_similar(SRC_SCHEMA    VARCHAR,
-                                                            SRC_TBL       VARCHAR,
-                                                            XCLUDE_COLS   VARCHAR,
-                                                            TRANS_COLS    VARCHAR,
-                                                            TRANS_ALIASES VARCHAR,
-                                                            TRG_SCHEMA    VARCHAR,
-                                                            TRG_TBL       VARCHAR)
+CREATE OR REPLACE PROCEDURE {{.output_schema}}.ctas_similar(
```
colmsnowplow: This is great - there's good reason to keep it as a stored procedure. As we realised when we discussed, it can be defined without any inputs since it's only used once - which should improve readability.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -334,42 +300,41 @@ CREATE OR REPLACE PROCEDURE {{.output_schema}}.commit_with_metadata(
...
+                                                 TRANS_ALIASES VARCHAR,
+                                                 TARGET_TABLE  VARCHAR,
+                                                 FROM_CLAUSE   VARCHAR,
+                                                 WHERE_CLAUSE  VARCHAR)
   RETURNS VARCHAR
   LANGUAGE JAVASCRIPT
   AS
   $$
 
+  var source_split = SOURCE_TABLE.split('.');
```
colmsnowplow: Nice way to simplify inputs

web/v1/snowflake/sql-runner/sql/standard/03-sessions/01-main/06-commit-sessions.sql
```diff
@@ -15,26 +15,37 @@
...
-    UPPER('{{.output_schema}}'),               -- target schema
-    UPPER('sessions{{.entropy}}'),             -- target table
-    UPPER('domain_sessionid'),                 -- join key
-    UPPER('start_tstamp'),                     -- partition key
-    {{or .upsert_lookback_days 30}}            -- upsert_lookback_days
+-- commit tables
+CALL {{.output_schema}}.commit_tables(
+  UPPER('{{.stage_next}}'),
+  'FALSE',
+  UPPER('{{.scratch_schema}}.sessions_userid_manifest_this_run{{.entropy}}'),
```
colmsnowplow: I'm confused about the manifest-y staged table being used in the same way as the staged tables. How does this work?

colmsnowplow:
Looking good. Remaining on checklist for this release:

- ~A few code comments~
- ~Resolve duplicates bug we discovered~
- ~Review of readmes etc.~ 

Update: All done.

.scripts/run_config.sh
```diff
@@ -37,16 +37,25 @@ cleanup() {
...
+  elif [ "${storage_target}" = "snowflake" ] ; then
+    PASSWORD=${SNOWFLAKE_PASSWORD:-$CREDENTIALS}
+  else
+    echo "Storage target mismatch!" >&2 ; exit 1;
+  fi
 
 elif [ "$target" == "BigQuery" ]; then
 
-  BIGQUERY_CREDS=${BIGQUERY_CREDS:-$CREDENTIALS}
+  PASSWORD=${BIGQUERY_CREDS:-$CREDENTIALS}
```
colmsnowplow: Using PASSWORD here breaks the sed later on - because the value isn't compatible for a simple substitution. It throws:

`sed: 1: "s/PASSWORD_PLACEHOLDER/ ...": bad flag in substitute command: 'c'`

(It's also not actually a password. It's a JSON of service account credentials for BQ)

Since the BQ playbooks don't need passwords, it's fine not to have any value in PASSWORD since that `sed` is just substituting nothing.

.scripts/run_config.sh
```diff
@@ -98,7 +107,7 @@ do
 
   # If printing sql to file, comment out metadata strings in the SQL
   if [ ! -z "$OUTPUT_PATH" ]; then
-    sed -E -i '' '/^([0-9]{4}\/(0[1-9]|1[0-2])\/(0[1-9]|[1-2][0-9]|3[0-1]) (2[0-3]|[01][0-9]):[0-5][0-9]:[0-5][0-9] Step name:|Query name: |Query path: )/ s/^/-- /' $OUTPUT_FILE
+    sed -E -i '/^([0-9]{4}\/(0[1-9]|1[0-2])\/(0[1-9]|[1-2][0-9]|3[0-1]) (2[0-3]|[01][0-9]):[0-5][0-9]:[0-5][0-9] Step name:|Query name: |Query path: )/ s/^/-- /' $OUTPUT_FILE
```
colmsnowplow: This change also breaks the intended behaviour - the intention of this line is to comment out the metadata that sql-runner produces with `fillTemplates` (below). Without the extra `''` it doesn't do this.

```
-- 2021/02/05 19:13:29 Step name: 01-page-views-join-setup
-- Query name: 01-page-views-join-setup
-- Query path: /Users/cog/GitHub/data-models/web/v1/bigquery/sql-runner/sql/custom/02-page-views-join/01-page-views-join-setup.sql
```

.scripts/run_config.sh
```diff
@@ -37,16 +37,25 @@ cleanup() {
 }
 trap cleanup EXIT
 
+mkdir -p "${root_path}/tmp"
+
 if [ $target == "Default" ]; then
 
-  PASSWORD=${DB_PASSWORD:-$CREDENTIALS}
+  storage_target=$(basename $(dirname "${model_path}"))
```
colmsnowplow, line 44: This is a good argument for the config supporting "Redshift" and "Snowflake" instead of/as well as "Default"

.scripts/run_config.sh
adatzer: Ok, thank you for trying it with bigquery! I will correct this so.

.scripts/run_config.sh
adatzer: Ok, for me it is the opposite happening. 
```
sed: can't read /^([0-9]{4}\/(0[1-9]|1[0-2])\/(0[1-9]|[1-2][0-9]|3[0-1]) (2[0-3]|[01][0-9]):[0-5][0-9]:[0-5][0-9] Step name:|Query name: |Query path: )/ s/^/-- /: No such file or directory
```

`sed (GNU sed) 4.7
Packaged by Debian`
But works as expected when removing.

web/v1/snowflake/sql-runner/sql/standard/03-sessions/01-main/06-commit-sessions.sql
adatzer: The code the procedure generates is the same e.g.
```
DELETE FROM SCRATCH.SESSIONS_USERID_MANIFEST_STAGED
WHERE DOMAIN_USERID IN (SELECT DOMAIN_USERID FROM SCRATCH.SESSIONS_USERID_MANIFEST_THIS_RUN); 
          
INSERT INTO SCRATCH.SESSIONS_USERID_MANIFEST_STAGED
  SELECT FROM SCRATCH.SESSIONS_USERID_MANIFEST_THIS_RUN;
```

.scripts/run_config.sh
paulboocock: You're not alone...
https://unix.stackexchange.com/questions/92895/how-can-i-achieve-portability-with-sed-i-in-place-editing

Using `perl` instead or not using `-i` seems like the best option.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
adatzer: @colmsnowplow just to confirm, could you please explain what do you expect to be the same? Do you maybe mean stage-joinKey Vs production-joinKey? Which 3 tables do you refer to?

web/v1/snowflake/sql-runner/sql/standard/03-sessions/01-main/06-commit-sessions.sql
colmsnowplow: Ok I see what you mean. I guess this is just another example of a convo we already had - wrapping both table updates in the one procedure makes it a bit confusing I think. :)

.scripts/run_config.sh
colmsnowplow: Ah it all makes sense now. OK I'm cool with either approach.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
colmsnowplow: So for example in the page views module we have three tables:

`page_views_this_run`, `page_views_staged`, `page_views`. These by definition have the exact same definition, including joinkey and partition key. 

So the function which handles their commit needs only one respective `paritionKey` and `joinKey` argument. If these are ever not the same, something is wrong and we shouldn't allow it.

web/v1/bigquery/README.md
```diff
@@ -23,7 +23,7 @@ Variables in each module's playbook can also optionally be configured also. See
 To run the entire standard model, end to end:
 
 ```bash
-bash .scripts/run_config.sh -b ~/pathTo/sql-runner -c web/v1/bigquery/configs/datamodeling.json;
+bash .scripts/run_config.sh -b ~/pathTo/sql-runner -c web/v1/bigquery/sql-runner/configs/datamodeling.json;
```
colmsnowplow: We can remove this from the PR - I'll update as part of BigQuery patch release.

.scripts/README.md
```diff
@@ -58,17 +58,17 @@ Note that this script does not enforce dependencies, rather runs the playbooks i
...
 
 # Runs the standard bigquery web model end to end.
 
-bash .scripts/run_config.sh -b ~/pathTo/sql-runner -c web/v1/bigquery/configs/datamodeling.json -d;
+bash .scripts/run_config.sh -b ~/pathTo/sql-runner -c web/v1/bigquery/sql-runner/configs/datamodeling.json -d;
 
 # Dry-runs the standard bigquery web model end to end.
 
-bash .scripts/run_config.sh -b ~/pathTo/sql-runner -c web/v1/bigquery/configs/example_with_custom.json -t -o tmp/sql;
+bash .scripts/run_config.sh -b ~/pathTo/sql-runner -c web/v1/bigquery/sql-runner/configs/example_with_custom.json -t .scripts/templates/bigquery.yml.tmpl -o output/sql;
```
colmsnowplow: My mistake here was to use `-t` instead of `-p`. So actually it should be:

`bash .scripts/run_config.sh -b ~/pathTo/sql-runner -c web/v1/bigquery/sql-runner/configs/example_with_custom.json -p -o output/sql`

README.md
```diff
@@ -73,7 +73,7 @@ For local use, the `.scripts/run_config.sh` script can be used to run a config -
 For those who wish to use other tools, one may configure playbooks and config JSON files for the desired model, then use the `.scripts/run_configs.sh` script's `-t` and `-o` flags to fill templates and output pure SQL to file:
 
 ```bash
-bash .scripts/run_config.sh -b ~/pathTo/sql-runner -c web/v1/bigquery/configs/example_with_custom.json -t -o tmp/sql;
+bash .scripts/run_config.sh -b ~/pathTo/sql-runner -c web/v1/bigquery/sql-runner/configs/example_with_custom.json -t .scripts/templates/bigquery.yml.tmpl -o tmp/sql;
```
colmsnowplow: Same as the scripts readme here and in the above paragraph  - I meant `-p` instead of `-t`

web/v1/snowflake/CHANGELOG
```diff
@@ -0,0 +1,5 @@
+Version 1.0.0 (2021-01-29)
+--------------------------
+Fix bug in credentials handling for pr_check and e2e scripts (#35)
```
colmsnowplow, line 3: This line will go into new top-level changelog (see bigquery 1.0.1 release). 

If this PR is ready first, then we can leave it as is and I'll change it in the other PR. :)

web/v1/snowflake/sql-runner/configs/XX_destroy_standard.json
```diff
@@ -0,0 +1,34 @@
+{
```
colmsnowplow: I prefer your naming here - but for the sake of consistency can we rename to `teardown_all.json`?

We should change them all to this in a follow-up though - could you create an issue for this?

web/v1/snowflake/sql-runner/playbooks/custom/02-page-views-join/01-page-views-join.yml.tmpl
colmsnowplow: Change of plan - we'll default to `atomic`, `scratch` and `derived` everywhere: https://github.com/snowplow/data-models/issues/41

web/v1/snowflake/sql-runner/sql/standard/00-setup/StoredProcedures.md
```diff
@@ -0,0 +1,84 @@
+# Stored procedures
```
colmsnowplow, line 1: I think we can remove this readme if commit_tables() can be simplified. If not, we can keep the readme but let's just mention `commit_tables()`, rather than explaining all the others (which we don't expect to be used).

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
adatzer: Thank you very much for explaining. This is indeed the case for the page-views module, but in sessions module there are 4 tables, where there is a difference:

- the join key for `sessions_this_run` - `sessions` is `domain_sessionid`
- the join key for `sessions_userid_manifest_this_run` - `sessions_userid_manifest_staged` is `domain_userid`

So for a procedure to abstract over all, different arguments seem needed.

web/v1/snowflake/sql-runner/sql/standard/00-setup/StoredProcedures.md
adatzer: Ok. What about `mk_transaction`?

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
colmsnowplow: Makes sense. Seems to be another reason to prefer this function to handle one update at a time.

web/v1/snowflake/sql-runner/sql/standard/00-setup/StoredProcedures.md
colmsnowplow: Do we expect it to be used directly? Seems to me that it's contained within `commit_table()`, no?

adatzer: No, it is not used within `commit_tables`. But it can be used in any upsert e.g. in custom modules.

colmsnowplow: Ok, I'm confused - when do we use it then?

Seems fine to leave it in but I'm confused about why it exists if we already provide a way to commit upserts

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
adatzer: Just to clarify here, which also relates to the comment on performance tests with splitting the transactions:
For both redshift and bigquery, those statements are inside a transaction. Maybe until this is changed there too (if not really needed) it makes sense to have the same logic for snowflake too, at least for v1.0.0. This was the reason for combining those statements in the same stored procedure (and this has also been the reason for many of the superficial differences so far).
Also, since Snowflake allows to overload a procedure name, usability is not really an issue in the long run.
I would prefer so instead of deviating at this point.

web/v1/snowflake/sql-runner/sql/standard/00-setup/StoredProcedures.md
adatzer: Sorry if i added confusion, maybe i am the one confused. `mk_transaction` is a way to wrap statements into a transaction. `commit_tables` does this for the commit step of the modules (including some more checks). So far, i indeed agree that `commit_tables` involves transaction logic but i consider them as different. Am i wrong here?

colmsnowplow: Yes I do see the difference between these - I think I phrased my question poorly. When do we need `mk_transaction`?

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
colmsnowplow: I'm not sure I've made the considerations v clear as I was still thinking about it when we spoke. Let me try to do so:

The best design for our functions (in this case a procedure) is the simplest. One unit of work at a time, and as generalisable as possible. 

We can see the compromise we've made here impact other parts of the model's design already (in other comments).

I think we also expect a lot of users to follow how to use these - most of them won't understand that the underlying models use two separate pre-production tables to stage updates, so it is not intuitive to them what all of the arguments are for.

These are more than just superficial considerations for me.

I see the decision here as a choice between sacrificing the best design, or sacrificing consistency in how transactions are done.

Nesting the transactions has very little (if any) actual impact on the model, as long as DELETE and INSERT are in the same transaction. We just do it in the other DBs because we can.

So for me, consistency in this aspect doesn't really outweigh the benefits of a simpler design.

Now, if performing two transactions does make a significant difference to performance, or if there is another material difference I'm missing, then we should definitely consider those things.

Does this make sense?

web/v1/snowflake/sql-runner/sql/standard/00-setup/StoredProcedures.md
adatzer: Currently `mk_transaction` is being used in: `98-base-manifest` and `98-manifest-and-truncate` of base and users module respectively and in `02-page-views-join-upsert` of the custom module.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
adatzer: > We can see the compromise we've made here impact other parts of the model's design already (in other comments).

It seems this is still under discussion.

> I see the decision here as a choice between sacrificing the best design, or sacrificing consistency in how transactions are done.

In my mind there is no need to sacrifice anything here. Snowflake allows for more options on this, so instead of one or the other, i see a potential to achieve both. Would you consider such options?

web/v1/snowflake/sql-runner/sql/standard/00-setup/StoredProcedures.md
colmsnowplow: Ah ok I see now. Ok cool, there's good reason to keep that too yeah.

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
colmsnowplow: Of course if we can avoid both problems I would prefer that. What solution do you propose?

adatzer:
_Address review changes III:_
1. Adding `users_limits` table to the users module
2. Remove `ctas_similar`, add `create_events_this_run`
3. Changes to the deduplication logic to address the bug we found as mentioned above. (`pr_check` passes). The deduplication logic is that of BigQuery, i.e. keep the first of duplicates. Back to using `page_view_id` as joinKey to upsert to `derived.page_views`
4. removed commit for issue 34
5. rename to `teardown_all.json`
6. split the transaction of committing to `staged` and production tables
7. change lockType to `"hard"`

web/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
adatzer: Thank you for considering this. So, the option is to define another stored procedure that does exactly as you propose, i.e. do one thing at the time. It can actually have the same name with just diff number of arguments, as Snowflake allows for this overload.
Advantages:
1. Not yet another procedure. Same name that allows for both use cases.
2. Keeping the same to other models' logic on the commit steps.
3. Easiest to change. The only thing is just to add this definition. Nothing else needs to change on the model, so the quickest at this point for v1.0.0

Disadvantages:
1. Some code duplication, as we overload.

I guess there may be more disadvantages as i don't have the experience to judge the quality of this design, but is indeed a quick change that at this point allows to have the simpler design you mention(and i agree) for any simpler use case and not having to sacrifice.
If you still think it is best design to do otherwise, i will go on to do so, but i estimate some more time will be needed.

colmsnowplow: I don't like duplicating the code, and I'm not sure if we're actually saving time by doing it this way - once you have a new function surely it's very little effort to use it in the SQL. 

However if you feel that's the best way forward then sure - we can clean up what we need to in a follow-up release.

adatzer: Thank you @colmsnowplow. I understand this overload option is not ideal, so i made the changes you suggest in the commit steps (the performance diff was around 3sec for 1,700,000 events(XSmall wh size), so i assume insignificant as argument to keep in same transaction). These changes were also force-pushed in "Address review changes III".

web/v1/snowflake/sql-runner/sql/standard/02-page-views/01-main/07-commit-page-views.sql
```diff
@@ -0,0 +1,64 @@
...
+    UPPER('{{.scratch_schema}}.page_views_staged{{.entropy}}'),    -- stage target
+    UPPER('page_view_id')                                          -- joinKey
+  );
+
+{{end}}
+
+-- production tables
+{{if ne (or .skip_derived false) true}}
+
+  CALL {{.output_schema}}.mk_transaction(
```
colmsnowplow, line 33: How come we don't use the procedure here?

adatzer: We use the `mk_transaction` one, because this is the only thing that happens there. Trying to wrap the one in terms of the other just to end up to the one we already have felt like the complexity that was before.

colmsnowplow: Ok cool. This approach is as valid as the other, so seems good to me. :)

web/v1/snowflake/CHANGELOG
colmsnowplow: Other PR is going after this release, so this comment need not apply.

web/v1/snowflake/sql-runner/sql/standard/04-users/01-main/02-users-sessions-this-run.sql
colmsnowplow: Sorry missed your last comment. Yes I think that's a good idea!

adatzer:
_Address review changes II_
(Contains fixing my mistakes on the scripts, as commit to be fixed up to the corresponding one referencing the scripts' issue. Compatibility to be confirmed.)

web/v1/snowflake/sql-runner/configs/datamodeling.json
colmsnowplow: I think it's worth changing this to `hard` - simply because if it's left on `soft` and deployed, that can generate a lot of annoying notifications.

.scripts/run_config.sh
colmsnowplow: I think we have made the changes here both to use `mkdir -p` and the env vars? If so we can resolve this.

.scripts/pr_check.sh
colmsnowplow: Same here - let's resolve if it's been done

web/v1/snowflake/sql-runner/sql/custom/02-page-views-join/XX-destroy/XX-destroy-page-views-join.sql
colmsnowplow: I think we changed this as well?

.scripts/run_config.sh
colmsnowplow: I see this has changed to `perl` I'll give it a test, but if it doesn't play nicely with Mac I can make an issue to fix it in the BQ release - no need to block this release for it.

colmsnowplow (approved):
LGTM. One v small change request to update the config to use hard lock. 

No need to wait to rebase, there's nothing left that should block release.

Once you've done so let me know - I can help w/the release process

paulboocock (approved):

PR #43, opened 2021-02-04
Release/web/bigquery/1.0.1

colmsnowplow:

PR #49, opened 2021-02-19
Release/web/bigquery/1.0.1

colmsnowplow:
New PR as scope changed from https://github.com/snowplow/data-models/pull/43

Few bugfixes and improved example.


web/v1/bigquery/sql-runner/sql/standard/01-base/01-main/03-sessions-to-process.sql
```diff
@@ -40,5 +40,13 @@ AS(
     AND a.domain_sessionid IS NOT NULL
     AND TIMESTAMP_DIFF(a.dvce_sent_tstamp, a.dvce_created_tstamp, DAY) <= {{or .days_late_allowed 3}}
     -- don't process data that's too late
+
+    {{if eq (or .derived_tstamp_partitioned false) true}}
```
paulboocock, line 44: Is it a bit odd that these become limited by both collector and derived when you toggle this on? Should it be one or the other?

web/v1/snowflake/CHANGELOG
```diff
@@ -1,4 +1,3 @@
 Version 1.0.0 (2021-02-11)
 --------------------------
-Fix bug in credentials handling for pr_check and e2e scripts (#35)
```
paulboocock, line 3: Why is this PR touching the snowflake changelog?

web/v1/snowflake/README.md
```diff
@@ -59,7 +59,7 @@ playbooks/
 To run the entire standard model, end to end:
 
 ```bash
-bash .scripts/run_config.sh -b ~/pathTo/sql-runner -c web/v1/snowflake/sql-runner/configs/datamodeling.json;
+bash .scripts/run_config.sh -b ~/pathTo/sql-runner -c web/v1/snowflake/sql-runner/configs/datamodeling.json -t .scripts/templates/snowflake.yml.tmpl;
```
paulboocock, line 62: This one ends in .yml.tmpl but the redshift one has no extension. Maybe the redshift one is missing the extension but I noticed it here so this is where my comment ended up :D

paulboocock (approved):

web/v1/bigquery/sql-runner/sql/standard/01-base/01-main/03-sessions-to-process.sql
colmsnowplow: Since the manifests and all logic later on depend on collector_tstamp, it'd be a lot more changes to just use derived. ( I'm not even sure derived can be reliably used alone, since it depends on device clock).

Am adding this change because it's a simple way to allow backwards compatibility for older customers who still partition by derived tstamp (default is collector tstamp now, and officially that's what the model supports). If it involves further change then it's not worth including at all IMO.

web/v1/snowflake/CHANGELOG
colmsnowplow: Because we introduce the CHANGELOG in the root of the repo, which is where we should record issues/commits that aren't specific to a given model. So we're moving the record of all those commits to the top changelog. Same for Redshift.

This does feel a bit of an awkward way to go about it, happy to change approach of course.

web/v1/snowflake/README.md
colmsnowplow: Good spot, thanks will fix.

web/v1/bigquery/sql-runner/sql/standard/01-base/01-main/03-sessions-to-process.sql
paulboocock: That seems fair to me!

web/v1/snowflake/CHANGELOG
paulboocock: I feel like we perhaps have diuscussed this before but does the root changelog contain ALL changes/releases? I think I'd be surprised it I opened it and only say "central" fixes in there.
I'm fine that the project specific change logs only include changes related to that release of that subproject though.

colmsnowplow: We did discuss briefly but I don't think we settled on this detail. 

Makes sense though, I'll change it - I was never v certain on this call. :)

CHANGELOG
```diff
@@ -0,0 +1,45 @@
+BigQuery Web Version 1.0.1 (2021-02-XX)
```
adatzer: Just to replace `XX` here.

web/v1/bigquery/CHANGELOG
```diff
@@ -1,7 +1,14 @@
+Version 1.0.1 (2021-02-XX)
```
adatzer: Just `XX` here too

CHANGELOG
colmsnowplow: Yup, left these as placeholders until all reviews are done. :)

web/v1/bigquery/sql-runner/sql/custom/02-custom-example/01-link-clicks.sql
```diff
@@ -0,0 +1,38 @@
+-- We use entropy templates to allow ourselves easier separation of dev and prod runs.
+CREATE OR REPLACE TABLE {{.scratch_schema}}.link_clicks{{.entropy}} AS(
+  WITH click_count AS(
+    SELECT
+      ev.page_view_id,
+      COUNT(DISTINCT ev.event_id) AS link_clicks,
```
adatzer: Just a question: Here it seems all events are counted per `page_view_id`, not only the `link_click`s. Is this the case?

colmsnowplow: Oh true. Good spot, thanks for flagging. Have pushed a fix!

adatzer (approved):
LGTM!

PR #55, opened 2021-03-09
Release/web/bigquery/1.0.2

colmsnowplow:
Introduces a refactor of how we pull in standard contexts for the page views module, via a stored procedure.

New stored procedure introduced which:

- Takes the prefix for a context column name (eg. `contexts_com_iab_snowplow_spiders_and_robots_1`)
- Checks table metadata for all field paths in that column
- Looks in only the first item in the array for that column
- Extracts all top-level fields excluding array or struct fields (these pose more complicated problems)
- Creates a scratch table from the data in events_staged, coalescing all columns - with the input string as the name

If the context in question is disabled in the config, then we create a table with null values.

This is then joined at the end of the page views step, on page_view_id and event_id. Since event_id is deduplicated in base (choosing first row), and page_view_id is deduplicated for page_view_events, then joining on both page_view_id and event_id gives us a 1:1 join.


web/v1/bigquery/sql-runner/sql/standard/00-setup/01-main/01-functions-and-procedures.sql
```diff
@@ -125,3 +125,43 @@ BEGIN
...
+    SELECT ARRAY_TO_STRING(
+      ARRAY_AGG(CONCAT('COALESCE(', ARRAY_TO_STRING(paths, ', '), ', NULL) AS ',  field_name)), ', ')
+    FROM(
+      -- Get field names and their corresponding paths
+      SELECT
+        SPLIT(field_path, '.')[SAFE_OFFSET(1)] AS field_name,
+        ARRAY_AGG(CONCAT(column_name, '[SAFE_OFFSET(0)].', SPLIT(field_path, '.')[SAFE_OFFSET(1)]) ORDER BY column_name DESC) AS paths
+
+      FROM {{.scratch_schema}}.INFORMATION_SCHEMA.COLUMN_FIELD_PATHS
+      WHERE table_name = 'events_staged'
```
bill-warner: Include entropy?

web/v1/bigquery/sql-runner/sql/standard/00-setup/01-main/01-functions-and-procedures.sql
```diff
@@ -125,3 +125,43 @@ BEGIN
...
+    RAISE USING MESSAGE = 'ERROR: Cannot combine context versions: No eligible top-level columns found.';
+
+  END IF;
+
+  -- Create scratch table with extracted data:
+  EXECUTE IMMEDIATE CONCAT("""CREATE OR REPLACE TABLE {{.scratch_schema}}.""",
+                            columns_prefix,
+                            """{{.entropy}} AS( SELECT event_id, page_view_id, collector_tstamp, derived_tstamp, """,
+                            COLUMN_COALESCE,
+                            """ FROM {{.scratch_schema}}.events_staged )""");
```
bill-warner: entropy?

bill-warner (commented):
Good work. Only spotted one issue around entropy suffix.

One thing I found when testing, should [this line](https://github.com/snowplow/data-models/blob/8ec695397a5d6f7ccfd89400c3f1254a93b84cf7/.scripts/run_config.sh#L69) be `cat` rather than `echo`? My assumption was that the env_var for BG creds would be a path to the credentials json rather than the raw credentials.

paulboocock (approved):
That's a really nice solution for BigQuery users. 
Is there a README to document this stored procedure? I see it being very useful for people building their own custom modules too.

colmsnowplow:
@paulboocock I figured because it's so limited, it would be better to implement improvements that allow it to work with struct and array fields first.

But on second thoughts I see no harm in adding some documentation and letting users have a go with what's already there.

web/v1/bigquery/sql-runner/sql/custom/README.md
```diff
@@ -100,6 +100,30 @@ derived.commit_table(
...
+
+The BigQuery loader for the Snowplow pipeline is unique in that it creates a new column for every new schema for an event or entity, rather than just for major version changes.
+
+The model uses a separate stored procedure to solve this problem internally, which might be useful for custom implementations too:
+
+```SQL
+CALL {{.output_schema}}.combine_context_versions('contexts_com_iab_snowplow_spiders_and_robots_1')
+```
+
+The `combine_context_versions` stored procedure will find all fields within across multiple versions of a context column, and create a new table which coalesces each top-level field, except for arrays or structs.
```
bill-warner: Minor but might read better as: 'The `combine_context_versions` stored procedure will find all fields within the context across multiple versions of a context column'

bill-warner (approved):

PR #70, opened 2021-04-01
Update images for accessibility (close #69)

### LINKED ISSUE #69: Update images for accessibility
As a colour blind user I have a hard time reading the diagrams, particularly the red boxes. I have to zoom in so the box fills most of the screen in order to read the text. Can the text be updated so it has a higher contrast (e.g. white text?) 

Thanks!

colmsnowplow:


paulboocock (approved):
Much better! They are significantly more readable now, the dark red with black text was particularly hard to read previously and they are much clearer now. The colour palette appears more appropriate for all too.

PR #74, opened 2021-04-14
Release/mobile/bigquery/1.0.0

bill-warner:


.scripts/run_config.sh
```diff
@@ -52,7 +52,7 @@ if [ $target == "Default" ]; then
 
 elif [ "$target" == "BigQuery" ]; then
 
-  BIGQUERY_CREDS=${BIGQUERY_CREDS:-$CREDENTIALS}
```
colmsnowplow, line 55: I'm on board with changing the behaviour here to point to a file rather than a blob of JSON, two comments on that:

- Let's make sure the readme is updated with it
- If we're doing this, we can just call the env var GOOGLE_APPLICATION_CREDENTIALS and remove this whole piece of logic :)

.test/great_expectations/expectations/mobile/v1/mobile_metadata.json
```diff
@@ -103,7 +103,8 @@
   "meta": {
     "versions": {
       "test_suite_version": "1.0.0",
```
colmsnowplow: Test suite has been updated, bump version to 1.0.1

mobile/v1/bigquery/sql-runner/sql/standard/00-setup/01-main/01-functions-and-procedures.sql
```diff
@@ -0,0 +1,305 @@
...
+                  'events',                 -- source_table
+                  '["session_id","session_index","previous_session_id","user_id","first_event_id"]', --source_fields. Quoted array to allow for concat
+                  'contexts_com_snowplowanalytics_snowplow_client_session_1_0', -- columns_prefix
+                  true,                     -- rename_y_n
+                  '["session_id","session_index","previous_session_id","device_user_id","session_first_event_id"]', -- rename_fields
+                  SESSION_CONTEXT_COLUMNS   -- returned context coalesce columns
+                  );
+
+  SET SESSION_ID = (
+    SELECT REGEXP_EXTRACT(SESSION_CONTEXT_COLUMNS, r"(COALESCE\(contexts_com_snowplowanalytics_snowplow_client_session_1_0_[0-9]\[SAFE_OFFSET\(0\)\]\.session_id.{1,}) AS session_id")
```
colmsnowplow: As we discussed on a catchup, I think the fact that this has to be done is a sign that we might organise our function better. It might be cleaner if we had a function which output an array of COALESCE statement strings, and then consumed that twice: once to concatenate them into what we use in the sql later on, and one to get the session ID one only.

mobile/v1/bigquery/sql-runner/sql/standard/00-setup/01-main/01-functions-and-procedures.sql
```diff
@@ -0,0 +1,305 @@
...
+      CAST(NULL AS STRING) AS open_idfa,
+      CAST(NULL AS STRING) AS network_technology,
+      CAST(NULL AS STRING) AS network_type""");
+
+  END IF;
+
+END;
+
+
+CREATE OR REPLACE PROCEDURE {{.output_schema}}.mobile_app_errors_fields (OUT APP_ERRORS_EVENTS_COLUMNS STRING)
```
colmsnowplow, line 278: As we mentioned on the call, it seems odd to have procedures that simply wrap other procedures. Worth considering whether there's a better way (perhaps we call combine_column_versions into a variable at the beginning of the step in which we use it, for example)

mobile/v1/bigquery/sql-runner/sql/standard/02-screen-views/01-main/00-setup-screen-views.sql
```diff
@@ -0,0 +1,225 @@
...
+  geo_region_name STRING,
+  geo_timezone STRING,
+  user_ipaddress STRING,
+  useragent STRING,
+  carrier STRING,
+  network_technology STRING,
+  network_type STRING,
+  build STRING,
+  version STRING,
+  load_tstamp TIMESTAMP
```
colmsnowplow: Nit: Something strikes me as off about the name `load_tstamp` - feels like it should be the time an event was loaded to database. Perhaps `model_tstamp` or `datamodel_tstamp`, or some other name might fit?

mobile/v1/bigquery/sql-runner/sql/standard/00-setup/01-main/01-functions-and-procedures.sql
```diff
@@ -0,0 +1,305 @@
...
+                    'contexts_com_snowplowanalytics_snowplow_mobile_context_1_0', -- columns_prefix
+                    false,                   -- rename_y_n
+                    '[""]',                  -- rename_fields
+                    MOBILE_CONTEXT_COLUMNS   -- returned context coalesce columns
+                    );
+
+  ELSE
+
+    SET MOBILE_CONTEXT_COLUMNS = (
+      """CAST(NULL AS STRING) AS device_manufacturer,
```
colmsnowplow, line 252: Feels like should need to do this often. Perhaps there's a way to make it a generic function (at some point in future).

mobile/v1/bigquery/sql-runner/sql/standard/03-optional-modules/01-app-errors/01-main/01-app-errors.sql
```diff
@@ -0,0 +1,111 @@
...
+        e.session_index,
+        e.previous_session_id,
+        e.session_first_event_id,
+
+        e.dvce_created_tstamp,
+        e.collector_tstamp,
+        e.derived_tstamp,
+
+        --Error details
+        %s,
```
colmsnowplow: Feels more future-proof to have this at the end of the query, since that allows us to handle additional fields in future without a table-definition headache.

mobile/v1/bigquery/sql-runner/sql/standard/05-users/01-main/04-users-aggs.sql
```diff
@@ -0,0 +1,58 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+CREATE OR REPLACE TABLE {{.scratch_schema}}.mobile_users_aggregates{{.entropy}}
+AS(
+
+  WITH prep AS (
```
colmsnowplow: No need for a subquery here.

.scripts/run_config.sh
colmsnowplow: Oh and also if we do that we should remove the bit that deletes the credentials file.

colmsnowplow:
@bill-warner I found an issue.

When I run the model with all the configurable contexts in base module set to false, it runs fine. When I then run it again with the mobile context enabled, commit table fails with:

> 2021/04/21 16:07:43 FAILURE: 08-commit-base (step 08-commit-base @ target ), ERROR: googleapi: Error 400: ERROR: Source table is missing column(s) which exist in target table. at [com-snplow-eng-gcp.derived_dev1.commit_table:23:5]

This is the error we manually throw in the commit table function when the column definitions are incompatible with automigrating.

This appears to only happen for the mobile context.

I am unsure as to why this happens, but I do have some suspicions. Let me know if you're struggling to map back to the source of the problem and we can rubber duck it.

mobile/v1/bigquery/sql-runner/sql/standard/00-setup/01-main/01-functions-and-procedures.sql
```diff
@@ -0,0 +1,305 @@
...
+      CAST(NULL AS STRING) AS device_model,
+      CAST(NULL AS STRING) AS os_type,
+      CAST(NULL AS STRING) AS os_version,
+      CAST(NULL AS STRING) AS android_idfa,
+      CAST(NULL AS STRING) AS apple_idfa,
+      CAST(NULL AS STRING) AS apple_idfv,
+      CAST(NULL AS STRING) AS carrier,
+      CAST(NULL AS STRING) AS open_idfa,
```
colmsnowplow, line 269: I think the bug I reported in the main comment might be because the procedure above doesn't list the columns in the same order as we do here. `commit_table` does enforce that column indexes match across source and target.

.test/great_expectations/expectations/mobile/v1/mobile_metadata.json
bill-warner: How has it changed? Adding GBQ to the expectation?

mobile/v1/bigquery/sql-runner/sql/standard/00-setup/01-main/01-functions-and-procedures.sql
bill-warner: Fair suggestion. I will see what I can do here.

mobile/v1/bigquery/sql-runner/sql/standard/02-screen-views/01-main/00-setup-screen-views.sql
bill-warner: Sure. I am easy on this one. Will change

mobile/v1/bigquery/sql-runner/sql/standard/03-optional-modules/01-app-errors/01-main/01-app-errors.sql
bill-warner: Makes sense to mitigate future pain. Deviates from the RS model then not the end of the world. Will change

mobile/v1/bigquery/sql-runner/sql/standard/00-setup/01-main/01-functions-and-procedures.sql
colmsnowplow: 👍  Like I said not something that's essential for this release, what you have will work. But worth bearing it in mind.

.test/great_expectations/expectations/mobile/v1/mobile_metadata.json
colmsnowplow: We discussed on the call but to clarify on the record - there are changes to the entire suite as a whole. I picked a bad file to put this comment on!

bill-warner:
Ok I think I have made all the discussed changes.

.scripts/run_test.sh
```diff
@@ -22,35 +22,10 @@ cd $root_path/.test
...
-  export GOOGLE_APPLICATION_CREDENTIALS=${GOOGLE_APPLICATION_CREDENTIALS:-$root_path/tmp/bq_creds.json}
-
-else
-
-  # If not BQ, take the relevant env var if it exists, set it to whatever's provided otherwise.
-  export REDSHIFT_PASSWORD=${REDSHIFT_PASSWORD:-$CREDENTIALS}
-  export SNOWFLAKE_PASSWORD=${SNOWFLAKE_PASSWORD:-$CREDENTIALS}
-
-fi
+# Take the relevant env var if it exists, set it to whatever's provided otherwise.
```
colmsnowplow, line 25: Removing all this code is quite satisfying, very glad this change came out of the review.

I think we should have a specific issue for this in the changelog as it's a change in behaviour - for when you rebase before release @bill-warner . :)

colmsnowplow (approved):
Can confirm the mobile context bug is resolved. I like the organisation of the functions better this way too. 

Looks ready for release to me! Great work.

PR #86, opened 2021-04-28
Add Snowflake mobile model v1 (Close #85)

### LINKED ISSUE #85: Add Snowflake mobile model v1

bill-warner:
Couple of points to call out:

- I have added a new procedure, `commit_table`, which replicates the same procedure in GBQ. This is in place of the `commit_staged` which was previously used in the Snowflake web model. I prefer this as it is a more general procedure that can also be used by the users for custom modules.
- I have also modified `column_check`. The `AUTOMIGRATE` input is now a BOOLEAN rather than a VARCHAR. This made more sense to me and again aligns with the GBQ model. The issue here is that this procedures is used in the web model too, and hasnt been updated in that model. However from what I understand Snowflake can have two procedures with the same name but different input types so in theory if the users is running both models the procedures shouldnt clash. Alternatively we could rename this to `column_check_v2` or something, or update the web model. 
- I removed the duplicates step `07-duplicates` in the base module. Given we never actually fully remove dupes, the insert overwrite procedure seemed overkill here just to count how many rows we removed. Again this aligns with GBQ.
- `06-events-this-run`: I removed the stored procedure from the web model that effectively replicates an `EXCEPT` statement in GBQ. Instead, I have chosen to leave in contexts cols like the page_view context. This simplified the logic and can potentially avoid headache later down the line if new fields are added to these columns that are later required.

### DIFF
--- a/mobile/v1/snowflake/README.md
+++ b/mobile/v1/snowflake/README.md
@@ -1,0 +1,240 @@
+# Snowflake v1 mobile model README
+
+This readme contains a quickstart guide, and details of how the modules interact with each other. For a guide to configuring each module, there is a README in each of the modules' `playbooks` directory.
+
+To customise the model, we recommend following the guidance found in the README in the `sql/custom` directory.
+
+## Quickstart
+
+### Prerequisites
+
+[SQL-runner 0.9.3](https://github.com/snowplow/sql-runner) or later must be installed, and a dataset of mobile events from either the Snowplow [iOS tracker](https://docs.snowplowanalytics.com/docs/collecting-data/collecting-from-own-applications/objective-c-tracker/) or [Android tracker](https://docs.snowplowanalytics.com/docs/collecting-data/collecting-from-own-applications/android-tracker/) must be available in the database. The session context and screen view events most both be enabled for the mobile model to run. 
+
+### Configuration
+
+#### Authentication
+
+First, fill in the connection details for the target database in the relevant template in `.scripts/template/snowflake.yml.tmpl`.
+
+Set an environment variable, `SNOWFLAKE_PASSWORD`, to your Snowflake password. See the README in `.scripts` for more detail.
+
+#### Contexts
+
+The following contexts can be enabled depending on your tracker configuration:
+
+- Mobile context
+- Geolocation context
+- Application context
+- Screen context
+
+By default they are disabled. For more details on how to enable please see the [README](sql-runner/playbooks/standard/01-base/README.md) in the Base module's playbooks folder.
+
+#### Optional Modules
+
+Currently the app errors module for crash reporting is the only optional module. More will be added in the future as the tracker's functionality expands.
+
+Assuming your tracker is capturing `application_error` events, the module can be enabled within the app errors playbook. For more details on how to enable please see the [README](sql-runner/playbooks/standard/03-optional-modules/01-app-errors/README.md) in the app errors module's playbooks folder.
+
+#### Variables
+
+Variables in each module's playbook can also optionally be configured also. See each playbook directory's README for more detail on configuration of each module.
+
+### Run using the `run_config.sh` script
+
+To run the entire standard model, end to end:
+
+```bash
+bash .scripts/run_config.sh -b ~/pathTo/sql-runner -c mobile/v1/snowflake/sql-runner/configs/datamodeling.json -t .scripts/templates/snowflake.yml.tmpl;
+```
+
+See the README in the `.scripts/` directory for more details.
+
+## Custom Modules
+
+A guide to creating custom modules can be found in the README of the `sql/custom/` directory of the relevant model. Each custom module created must consist of a set of sql files and a playbook, or set of playbooks. The helper scripts described above can also be used to run custom modules.
+
+## Testing
+
+### Setup
+
+Python3 is required.
+
+Install Great Expectations and dependencies, and configure a datasource:
+
+```bash
+cd .test
+pip3 install -r requirements.txt
+great_expectations datasource new
+```
+
+Follow the CLI guide to configure access to your database. The configuration for your datasource will be generated in `.test/great_expectations/config/config_variables.tml` - these values can be replaced by environment variables if desired.
+
+Please be aware that the names of the tables to test have been hardcoded in the [validation configs](.test/great_expectations/validation_configs). If you are using a custom values for any of the `entropy`, `scratch_schema` or `output_schema` variables within your playbooks, you will need to manually ammend the validation configs accordingly.
+
+If you have enabled any optional modules within the main mobile model, you will need to enable tests on these modules too. For more details on how to enable please see the [README](sql-runner/playbooks/tests/00-staging-reconciliation/README.md) in the staging reconciliation module's playbooks folder.
+
+### Using the helper scripts
+
+To run the test suites alone:
+
+```bash
+bash .scripts/run_test.sh -d snowflake -c perm_tables -m mobile -a {credentials (optional)}
+bash .scripts/run_test.sh -d snowflake -c temp_tables -m mobile -a {credentials (optional)}
+```
+
+To run an entire run of the standard model, and tests end to end:
+
+```bash
+bash .scripts/e2e.sh -b {path_to_sql_runner} -d snowflake -m mobile -a {credentials (optional)}
+```
+
+To run a full battery of ten runs of the standard model, and tests:
+
+```bash
+bash .scripts/pr_check.sh -b {path_to_sql_runner} -d snowflake -m mobile -a {credentials (optional)}
+```
+
+### Adding to tests
+
+Check out the [Great Expectations documentation](https://docs.greatexpectations.io/en/latest/) for guidance on using it to run existing test suites directly, create new expectations, use the profiler, and autogenerate data documentation.
+
+Quickstart to create a new test suite:
+
+`great_expectations suite new`
+
+## Modules detail
+
+### 01-base
+
+Inputs:             atomic tables, `{{.output_schema}}.mobile_base_event_id_manifest`, `{{.output_schema}}.mobile_base_session_id_manifest`
+
+Persistent Outputs: `{{.scratch_schema}}.mobile_events_staged`,
+
+Temporary Outputs:  `{{.scratch_schema}}.mobile_events_this_run`, `{{.scratch_schema}}.mobile_base_duplicates_this_run`
+
+The base module executes the incremental logic of the model - it retrieves all events for sessions with new data, deduplicates, and adds any enabled contexts.
+
+The base module's 'complete' playbook (`99-base-complete.yml.tmpl`) updates the two relevant manifests, and cleans up temporary tables. The lifecycle of the `{{.scratch_schema}}.mobile_events_staged` table is completed by the `99-sessions-complete.yml.tmpl` step of the sessions module, when the table is truncated. This truncation can only occur during the completion step of the sessions module as `{{.scratch_schema}}.mobile_events_staged` is required as an input to the sessions module. This differs to the web model where the page views module's complete step would contain the truncation step.
+
+The `{{.scratch_schema}}.mobile_events_this_run` table contains all events relevant only to this run of the model (since the last time the `99-base-complete.yml.tmpl` playbook has run). This table is dropped and recomputed _every_ run of the module, regardless of whether another module has used the data.
+
+If there is a requirement that a custom module consumes data _more frequently than the screen views module for example_, the `{{.scratch_schema}}.mobile_events_this_run` table may be used for this purpose.
+
+The `{{.scratch_schema}}.mobile_events_staged` table is incrementally updated to contain all events relevant to any run of the base module _since the last time the sessions module consumed it_ (ie since the last time the `99-sessions-complete.yml.tmpl` has run). This allows one to run the base module more frequently than the subsequent modules (if, for example, a custom module reads from events_this_run). 
+
+Detail on configuring the base module's playbook can be found [in the relevant playbook directory's README](sql-runner/playbooks/standard/01-base).
+
+### 02-screen-views
+
+Inputs:             atomic tables, `{{.scratch_schema}}.mobile_events_staged`
+
+Persistent Outputs: `{{.output_schema}}.mobile_screen_views`, `{{.scratch_schema}}.mobile_screen_views_staged`
+
+Temporary Outputs:  `{{.scratch_schema}}.mobile_screen_views_this_run`
+
+The screen views module takes `{{.scratch_schema}}.mobile_events_staged` as its input, joins in and deduplicates screen_view_id, calculates the standard mobile screen views model, and updates the production mobile_screen_views table. It also produces the `{{.scratch_schema}}.mobile_screen_views_staged` and `{{.scratch_schema}}.mobile_screen_views_this_run` tables.
+
+The screen views module's 'complete' playbook `99-screen-views-complete.yml.tmpl` cleans up temporary tables. The lifecycle of the `{{.scratch_schema}}.mobile_screen_views_staged` table is completed by the `99-sessions-complete.yml.tmpl` step (of the subsequent module).
+
+The `{{.scratch_schema}}.mobile_screen_views_this_run` table contains all events relevant only to this run of the model (since the last time the `99-screen-views-complete.yml.tmpl` playbook has run). This table is dropped and recomputed _every_ run of the module, regardless of whether another module has used the data.
+
+If there is a requirement that a custom module consumes data _more frequently than the sessions module_, the `{{.scratch_schema}}.mobile_screen_views_this_run` table may be used for this purpose.
+
+The `{{.scratch_schema}}.mobile_screen_views_staged` table is incrementally updated to contain all events relevant to any run of the screen views module _since the last time the sessions module consumed it_ (ie since the last time the `99-sessions-complete.yml.tmpl` playbook has run). This allows one to run the screen views module more frequently than the sessions module (if, for example, a custom module reads from mobile_screen_views_this_run).
+
+Detail on configuring the screen views module's playbook can be found [in the relevant playbook directory's README](sql-runner/playbooks/02-screen-views).
+
+### 03-optional-modules
+
+#### 01-app-errors
+
+Inputs:             atomic tables, `{{.scratch_schema}}.mobile_events_staged`
+
+Persistent Outputs: `{{.output_schema}}.mobile_app_errors`, `{{.scratch_schema}}.mobile_app_errors_staged`
+
+Temporary Outputs:  `{{.scratch_schema}}.mobile_app_errors_this_run`
+
+The app errors module takes `{{.scratch_schema}}.mobile_events_staged` as its input, joins in the app errors context, calculates the app errors model, and updates the production mobile_app_errors table. It also produces the `{{.scratch_schema}}.mobile_app_errors_staged` and `{{.scratch_schema}}.mobile_app_errors_this_run` tables.
+
+This crash reporting module is disabled by default since it is not a requirement to run the mobile model. Despite this, the `{{.scratch_schema}}.mobile_app_errors_staged` table will be created irrespectively. This is to allow the sessions module to run correctly where the `{{.scratch_schema}}.mobile_app_errors_staged` table is required as an input.
+
+The app errors module's 'complete' playbook `99-app-errors-complete.yml.tmpl` cleans up temporary tables. The lifecycle of the `{{.scratch_schema}}.mobile_app_errors_staged` table is completed by the `99-sessions-complete.yml.tmpl` step (of the subsequent module).
+
+The `{{.scratch_schema}}.mobile_app_errors_this_run` table contains all events relevant only to this run of the model (since the last time the `99-app-errors-complete.yml.tmpl` playbook has run). This table is dropped and recomputed _every_ run of the module, regardless of whether another module has used the data.
+
+If there is a requirement that a custom module consumes data _more frequently than the sessions module_, the `{{.scratch_schema}}.mobile_app_errors_this_run` table may be used for this purpose.
+
+The `{{.scratch_schema}}.mobile_app_errors_staged` table is incrementally updated to contain all events relevant to any run of the screen views module _since the last time the sessions module consumed it_ (ie since the last time the `99-sessions-complete.yml.tmpl` playbook has run). This allows one to run the app errors module more frequently than the sessions module (if, for example, a custom module reads from mobile_app_errors_this_run).
+
+Detail on configuring the app errors module's playbook can be found [in the relevant playbook directory's README](sql-runner/playbooks/03-optional-modules/01-app-errors).
+
+### 04-sessions
+
+Inputs:             `{{.scratch_schema}}.mobile_screen_views_staged`, `{{.scratch_schema}}.mobile_app_errors_staged`, `{{.scratch_schema}}.mobile_events_staged`
+
+Persistent Outputs: `{{.output_schema}}.mobile_sessions`, `{{.scratch_schema}}.mobile_sessions_userid_manifest_staged`
+
+Temporary Outputs:  `{{.scratch_schema}}.mobile_sessions_this_run`
+
+The sessions module takes the `_staged` output tables of the upstream modules as its input, calculates the standard sessions model, and updates the production sessions table. It also produces the `{{.scratch_schema}}.mobile_sessions_userid_manifest_staged` and `{{.scratch_schema}}.mobile_sessions_this_run{{.entropy}}` tables.
+
+Unlike the other modules, the sessions module outputs a manifest of IDs as its staged table rather than a table containing all unprocessed data - this is due to the fact that the users step requires a longer lookback than the incremental structure contains, so there are obviously efficiency limitations.
+
+The sessions module's 'complete' playbook `99-sessions-complete.yml.tmpl` truncates the input tables, and cleans up temporary tables. The lifecycle of the `{{.scratch_schema}}.mobile_sessions_userid_manifest_staged` table is completed by the `99-users-complete.yml.tmpl` step (of the subsequent module).
+
+The `{{.scratch_schema}}.mobile_sessions_this_run` table contains all events relevant only to this run of the model (since the last time the `99-sessions-complete.yml.tmpl` playbook has run). This table is dropped and recomputed _every_ run of the module, regardless of whether another module has used the data.
+
+If there is a requirement that a custom module consumes data _more frequently than the users module_, the `{{.scratch_schema}}.mobile_sessions_this_run` table may be used for this purpose.
+
+The `{{.scratch_schema}}.mobile_sessions_userid_manifest_staged` table is incrementally updated to contain all IDs relevant to any run of the sessions module _since the last time the users module consumed it_ (ie since the last time the `99-users-complete.yml.tmpl` playbook has run). This allows one to run the sessions module more frequently than the users module (if, for example, a custom module reads from sessions_this_run and is more frequent than the page views module).
+
+Detail on configuring the sessions module's playbook can be found [in the relevant playbook directory's README](sql-runner/playbooks/04-sessions).
+
+### 05-users
+
+Inputs:             `{{.scratch_schema}}.mobile_sessions_userid_manifest_staged`, `{{.output_schema}}.mobile_users_manifest`
+
+Persistent Outputs: `{{.output_schema}}.mobile_users`
+
+Temporary Outputs:  `{{.scratch_schema}}.mobile_users_this_run`
+
+The sessions module takes `{{.scratch_schema}}.mobile_sessions_userid_manifest_staged` as its input, alongside the `{{.output_schema}}.mobile_users_manifest` table (which is self-maintained within the users module). It calculates the standard users model, and updates the production users table. It also produces the `{{.scratch_schema}}.mobile_users_this_run` table.
+
+Unlike the other modules, the users module doesn't take an input that contains all information required to run the module. It uses the `{{.output_schema}}.mobile_users_manifest` table to manage efficiency, and queries the sessions table to process data as far back in history as is required.
+
+The users module's 'complete' playbook `99-users-complete.yml.tmpl` truncates the `{{.scratch_schema}}.mobile_sessions_userid_manifest_staged` table, commits to the `{{.output_schema}}.mobile_users_manifest` and cleans up temporary tables. There is no `_staged` table for this module, as there are no subsequent modules.
+
+The `{{.scratch_schema}}.mobile_users_this_run` table contains all events relevant only to this run of the model (since the last time the `99-users-complete.yml.tmpl` playbook has run). This table is dropped and recomputed _every_ run of the module, regardless of whether another module has used the data.
+
+Detail on configuring the users module's playbook can be found [in the relevant playbook directory's README](sql-runner/playbooks/standard/05-users).
+
+## Scheduling
+
+### Asynchronous Runs
+
+While the model is configured by default to run the entire way through, i.e. from the base module through to the users module, it is possible to run each module independently. For instance one could run the screen views module hourly while only running the sessions module daily. To do so you should run hourly all modules up to and including the desired module i.e. the base and screen view modules. The sessions module can then be run on a daily schedule. A few points to note:
+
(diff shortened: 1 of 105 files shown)


mobile/v1/snowflake/sql-runner/sql/custom/04-session-goals/01-session-goals-staged.sql
```diff
@@ -0,0 +1,35 @@
+-- 1. Aggregate with a drop and recompute logic
```
adatzer, line 17: Do we add copyright to custom module files?

mobile/v1/snowflake/sql-runner/sql/custom/04-session-goals/02-session-goals-upsert.sql
```diff
@@ -0,0 +1,11 @@
+-- 2. Commit table procedure handles committing to prod, including table creation, and creation of new columns if 'automigrate' is set to TRUE
```
adatzer, line 17: add copyright here too?

mobile/v1/snowflake/sql-runner/sql/custom/04-session-goals/99-session-goals-cleanup.sql
```diff
@@ -0,0 +1,3 @@
+
```
adatzer, line 3: add copyright here too?

mobile/v1/snowflake/sql-runner/sql/custom/04-session-goals/99-session-goals-cleanup.sql
```diff
@@ -0,0 +1,3 @@
...
+
+DROP TABLE IF EXISTS {{.scratch_schema}}.session_goals_staged{{.entropy}};
```
adatzer, line 17: Why do we drop the staged table here? Would it make sense to add an `XX-destroy` for custom module?
Even though there is no module consuming this staged table, maybe dropping `staged` tables is normally part of destroy steps?

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,245 @@
...
+  Since ALTER TABLE is a DDL statement, it will be its own transaction.
+  This means that even if it was placed inside another transaction,
+    it would commit that, and then start another(implicit) for its own execution.
+  So, also it cannot be explicitly rolled back.
+  Input:
+  SOURCE_SCHEMA:  the schema of the source table
+  SOURCE_TABLE:   the source table name
+  TARGET_SCHEMA:  the schema of the target table
+  TARGET_TABLE:   the target table name
+  AUTOMIGRATE:    whether target tabled gets altered if needed (only 'TRUE' enables)
```
adatzer: Since `AUTOMIGRATE` is now a boolean, the `(only 'TRUE' enables)` part of the comment could be removed.

mobile/v1/snowflake/sql-runner/sql/standard/01-base/01-main/00-setup-base.sql
```diff
@@ -0,0 +1,271 @@
...
+
+    mkt_medium                  VARCHAR(255),
+    mkt_source                  VARCHAR(255),
+    mkt_term                    VARCHAR(255),
+    mkt_content                 VARCHAR(500),
+    mkt_campaign                VARCHAR(255),
+
+    se_category                 VARCHAR(1000),
+    se_action                   VARCHAR(1000),
+    se_label                    VARCHAR(1000),
```
adatzer: Would it make sense to make the length of `se_label` configurable? 
Or somehow support this coming change (https://github.com/snowplow-incubator/snowplow-snowflake-loader/issues/134)?

mobile/v1/snowflake/README.md
```diff
@@ -0,0 +1,240 @@
...
+
+This readme contains a quickstart guide, and details of how the modules interact with each other. For a guide to configuring each module, there is a README in each of the modules' `playbooks` directory.
+
+To customise the model, we recommend following the guidance found in the README in the `sql/custom` directory.
+
+## Quickstart
+
+### Prerequisites
+
+[SQL-runner](https://github.com/snowplow/sql-runner) must be installed, and a dataset of mobile events from either the Snowplow [iOS tracker](https://docs.snowplowanalytics.com/docs/collecting-data/collecting-from-own-applications/objective-c-tracker/) or [Android tracker](https://docs.snowplowanalytics.com/docs/collecting-data/collecting-from-own-applications/android-tracker/) must be available in the database. The session context and screen view events most both be enabled for the mobile model to run.
```
adatzer: Maybe add a note that minimum SQL Runner version required is 0.9.3?

mobile/v1/snowflake/sql-runner/playbooks/standard/00-setup/99-metadata-complete.yml.tmpl
```diff
@@ -0,0 +1,20 @@
...
+  :account:
+  :database:
+  :warehouse:
+  :username:
+  :password:
+:variables:
+  :model_version:      snowflake/mobile/1.0.0
+  :model:              mobile
+  :scratch_schema:     scratch
+  :output_schema:      derived
```
adatzer, line 13: Here too, could the `output_schema` be removed?

mobile/v1/snowflake/sql-runner/playbooks/custom/04-session-goals/99-session-goals-complete.yml.tmpl
```diff
@@ -0,0 +1,20 @@
...
+  :account:
+  :database:
+  :warehouse:
+  :username:
+  :password:
+:variables:
+  :model_version:           snowflake/mobile/1.0.0
+  :model:                   mobile
+  :scratch_schema:          scratch
+  :output_schema:           derived
```
adatzer: Could the `model` and `output_schema` variables be removed?

mobile/v1/snowflake/sql-runner/playbooks/standard/02-screen-views/01-screen-views-main.yml.tmpl
```diff
@@ -0,0 +1,44 @@
...
+    - :name: 02-screen-views-metadata
+      :file: standard/02-screen-views/01-main/02-screen-views-metadata.sql
+      :template: true
+- :name: 03-commit-screen-views
+  :queries:
+    - :name: 03-commit-screen-views
+      :file: standard/02-screen-views/01-main/03-commit-screen-views.sql
+      :template: true
+
+
```
adatzer: nit, empty lines

mobile/v1/snowflake/sql-runner/playbooks/standard/03-optional-modules/01-app-errors/01-app-errors-main.yml.tmpl
```diff
@@ -0,0 +1,45 @@
...
+    - :name: 02-app-errors-metadata
+      :file: standard/03-optional-modules/01-app-errors/01-main/02-app-errors-metadata.sql
+      :template: true
+- :name: 03-commit-app-errors
+  :queries:
+    - :name: 03-commit-app-errors
+      :file: standard/03-optional-modules/01-app-errors/01-main/03-commit-app-errors.sql
+      :template: true
+
+
```
adatzer: nit, empty lines

mobile/v1/snowflake/sql-runner/playbooks/standard/03-optional-modules/01-app-errors/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+### 01-app-errors-main
+
+`:input_schema:`       name of atomic schema
+
+`:scratch_schema:`     name of scratch schema  
+
+`:output_schema:`      name of derived schema
+
+`:enabled:`      			 boolean - Toggles the module on/off.
```
adatzer: nit, extra space

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,245 @@
...
+      snowflake.createStatement({sqlText: tgr_insert_stmt}).execute();
+      snowflake.createStatement({sqlText: `COMMIT;`}).execute();
+
+  } catch(ERROR) {
+
+      snowflake.createStatement({sqlText: `ROLLBACK;`}).execute();
+      throw ERROR;
+
+  }
+  return "ok. commit_staged succeeded.";
```
adatzer: just a typo: `commit_table` instead of `commit_staged`

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,245 @@
...
+      var cols_check = `CALL {{.output_schema}}.column_check(
+                            '` + SOURCE_SCHEMA + `'
+                          , '` + SOURCE_TABLE + `'
+                          , '` + TARGET_SCHEMA + `'
+                          , '` + TARGET_TABLE + `'
+                          , `  + AUTOMIGRATE.toString() + `);`;
+      snowflake.createStatement({sqlText: cols_check}).execute();
+  }
+
+  /* Prepare delete/insert statements */
```
adatzer, line 274: If `AUTOMIGRATE` is set to `FALSE` at the moment the `column_check` is not called. 
Since `AUTOMIGRATE` is passed to `column_check` as a variable, would it make sense to move the call to `column_check` out of the `if` block, so that it runs even if `AUTOMIGRATE` is set to false?
The benefit could be a more helpful error (coming from `column_check` and possibly pointing to automigration status), since at the moment we will only get errors straight from the execution of delete/inserts.

mobile/v1/snowflake/sql-runner/playbooks/custom/04-session-goals/01-session-goals-main.yml.tmpl
```diff
@@ -0,0 +1,25 @@
...
+- :name:
+  :type:      snowflake
+  :account:
+  :database:
+  :warehouse:
+  :username:
+  :password:
+:variables:
+  :model_version:           snowflake/mobile/1.0.0
+  :model:                   mobile
```
adatzer: Could the `model` variable be removed?

colmsnowplow:
> I have also modified column_check. The AUTOMIGRATE input is now a BOOLEAN rather than a VARCHAR. This made more sense to me and again aligns with the GBQ model. The issue here is that this procedures is used in the web model too, and hasnt been updated in that model. However from what I understand Snowflake can have two procedures with the same name but different input types so in theory if the users is running both models the procedures shouldnt clash. Alternatively we could rename this to column_check_v2 or something, or update the web model.

For sure we should update the web model when we can. Even though it's allowed I don't like having two distinct functions with the same name - especially since normal use of this design pattern isn't really what we're doing with it. 

I think raise an issue and see if we can get the change scheduled for the web model sooner rather than later.

colmsnowplow:
> 06-events-this-run: I removed the stored procedure from the web model that effectively replicates an EXCEPT statement in GBQ. Instead, I have chosen to leave in contexts cols like the page_view context. This simplified the logic and can potentially avoid headache later down the line if new fields are added to these columns that are later required.

So does this basically mean a breaking change to the web model when we integrate the base module changes into that?

bill-warner:
> > 06-events-this-run: I removed the stored procedure from the web model that effectively replicates an EXCEPT statement in GBQ. Instead, I have chosen to leave in contexts cols like the page_view context. This simplified the logic and can potentially avoid headache later down the line if new fields are added to these columns that are later required.
> 
> So does this basically mean a breaking change to the web model when we integrate the base module changes into that?

Technically not because the new base module will output to `web_events_staged` rather then `events_staged`.

bill-warner:
> > I have also modified column_check. The AUTOMIGRATE input is now a BOOLEAN rather than a VARCHAR. This made more sense to me and again aligns with the GBQ model. The issue here is that this procedures is used in the web model too, and hasnt been updated in that model. However from what I understand Snowflake can have two procedures with the same name but different input types so in theory if the users is running both models the procedures shouldnt clash. Alternatively we could rename this to column_check_v2 or something, or update the web model.
> 
> For sure we should update the web model when we can. Even though it's allowed I don't like having two distinct functions with the same name - especially since normal use of this design pattern isn't really what we're doing with it.
> 
> I think raise an issue and see if we can get the change scheduled for the web model sooner rather than later.

Ok would you suggest keeping the function naming as is and just align the web model ASAP? The web model release could be a good opportunity to release the integrated base module too.

adatzer (approved):
LGTM! Great work @bill-warner !

Indeed `AUTOMIGRATE` as a boolean makes much more sense and it is indeed possible to overload the procedures' names in Snowflake, so there is no clash until it is changed in the web model too.
On duplicates, even though their number could still have some value for users (could it indicate a problem?), i do see the point that it may actually be an overkill.

For reference, all tests i ran have succeeded.

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,245 @@
...
+  SOURCE_TABLE:          the source table to select data from
+  TARGET_SCHEMA:         the schema of the target table
+  TARGET_TABLE:          the target table to delete/insert into
+  JOIN_KEY:              the join key between the source and target tables
+  PARTITION_KEY:         the partition of the target table
+  AUTOMIGRATE:           whether to automigrate staged table
+  To drop:
+  DROP PROCEDURE {{.output_schema}}.commit_table(VARCHAR,VARCHAR,VARCHAR,VARCHAR,VARCHAR,VARCHAR,BOOLEAN);
+*/
+CREATE OR REPLACE PROCEDURE {{.output_schema}}.commit_table(SOURCE_SCHEMA VARCHAR,
```
colmsnowplow, line 227: We should indeed rename this function - since we have two distinct functions with the same arguments of different types, this isn't a safe way to use overloading.

Snowflake will cast the argument if it's the wrong type, which will cause unexpected behaviour: https://docs.snowflake.com/en/sql-reference/stored-procedures-usage.html#overloading-stored-procedure-names

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,245 @@
...
+  } catch(ERROR) {
+
+      snowflake.createStatement({sqlText: `ROLLBACK;`}).execute();
+      throw ERROR;
+
+  }
+  return "ok. commit_staged succeeded.";
+
+  // == Helpers ==
+  function list_cols(sch,tbl) {
```
colmsnowplow: Nit: Not something that has to change this release, but it would be preferable to define our helper functions independently, if possible. An example of why is that currently, we have no way to independently test that `list_cols()` behaves as expected.

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,245 @@
...
+  To drop:
+  DROP PROCEDURE {{.output_schema}}.mk_transaction(VARCHAR);
+*/
+CREATE OR REPLACE PROCEDURE {{.output_schema}}.mk_transaction(DML_STATEMENTS VARCHAR)
+  RETURNS VARCHAR
+  LANGUAGE JAVASCRIPT
+  AS
+  $$
+
+  dmls = DML_STATEMENTS.split(';')
```
colmsnowplow: Is there a reason to split the statements? I'm not seeing any benefit and it intuitively is better to just execute the SQL we are provided rather than arbitrarily splitting it into separate statements. 

If there are a set of related statements which include the delimiter `;`, then `mk_transaction()` won't be able to run them.

Not a blocker for this release, but if there isn't a reason to do it we should look to remove this in future.

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,245 @@
...
+  if (targetColumns.some(notIncludedIn(sourceColumns)) === true) {
+
+      throw "ERROR: Source table is missing column(s) which exist in target table.";
+
+  } else {
+
+      var columnAdditions = sourceColumns.filter(notIncludedIn(targetColumns));
+
+      if (sourceColumns.some(notIncludedIn(targetColumns)) === true) {
+          if (!AUTOMIGRATE) {
```
colmsnowplow: Nit: It's counterintuitive to call automigrate here. `column_check()` intuitively should be an atomic function which performs a column check, nothing else. Automigrate should be a feature of `commit_table()`. 

The fact that we have a conditional on automigrate in commit_table as well supports that design.

Another one that's not a blocker for this release, but if it won't delay us too much would be a cleaner design to manage moving forwards.

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,245 @@
...
+                                                            SOURCE_TABLE  VARCHAR,
+                                                            TARGET_SCHEMA VARCHAR,
+                                                            TARGET_TABLE  VARCHAR,
+                                                            AUTOMIGRATE   BOOLEAN)
+  RETURNS VARCHAR
+  LANGUAGE JAVASCRIPT
+  AS
+  $$
+
+  var delim = '~';
```
colmsnowplow: Nit: Feels like this should take an array rather than a string with an arbitrary delimiter.

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,245 @@
...
+      var columnAdditions = sourceColumns.filter(notIncludedIn(targetColumns));
+
+      if (sourceColumns.some(notIncludedIn(targetColumns)) === true) {
+          if (!AUTOMIGRATE) {
+              throw "ERROR: Target table is missing column(s),but automigrate is not enabled.";
+          } else {
+              // enforce order
+              for (var i = 0; i < targetColumns.length; i++) {
+                  if (targetColumns[i] !== sourceColumns[i]) {
+                      throw "ERROR: Can only migrate extra columns of the end of source"
```
colmsnowplow: Nit: Error message could be clearer

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,245 @@
...
+                  }
+              }
+              add_columns_to(TARGET_SCHEMA, TARGET_TABLE, columnAdditions.join(','));
+              return "ok.Columns added."
```
colmsnowplow: Nit: Being picky here but 'ok' in the success messages (throughout the procedures) seems colloquial and not necessary.

mobile/v1/snowflake/sql-runner/sql/standard/01-base/01-main/00-setup-base.sql
```diff
@@ -0,0 +1,271 @@
...
+AS (
+  SELECT
+    'seed'::VARCHAR(36) AS session_id,
+    '{{.start_date}}'::TIMESTAMP_NTZ AS min_tstamp
+);
+
+
+-- Create staged table
+-- Not defining mobile_events_staged as table likely to evolve. Suggest we remove web def below in future releases of SF model.
+{{if eq .model "web"}}
```
colmsnowplow: Why do we have an explicit table definition for web but not mobile?

colmsnowplow (approved):
LGTM, great work @bill-warner 

I left some nitpicks which can be deferred to another release. 

Approving on the condition that we address the one important change - renaming the overloaded `commit_table()` procedure. :)

colmsnowplow:
> Ok would you suggest keeping the function naming as is and just align the web model ASAP? The web model release could be a good opportunity to release the integrated base module too.

Addressed it in my review, but to leave nothing unanswered - we should both rename it and align the web model. The fact that 
a) a commit_table exists already,
 and that 
b) Snowflake will cast a wrong type instead of failing 
converge to leave us in an unsafe position without renaming it. 

In reality we likely won't hit this problem, since they happen to do analogous things, but this is by coincidence rather than design, so we're better off to remove scope for that problem ever manifesting itself.

mobile/v1/snowflake/sql-runner/sql/standard/01-base/01-main/00-setup-base.sql
bill-warner: Ultimately the staged table is going to be mutated by the `commit_table` function as automigrate = true so it seemed redundant to define the staging table. Personally I think it makes sense to remove the web definition. Thoughts?

colmsnowplow: On one hand, it can be useful to have the base fields defined somewhere for dev accessibility purposes. On the other, you're right that it is indeed redundant to do so. I'm happy with either call, I just think it should be consistent. :)

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
bill-warner: Will look to strip out the helper functions into their own functions. On this one specifically, I am not convinced we need it at all

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
bill-warner: I believe we have to be split as Snowflake does not support running multiple sql statements via `execute()`. We could supply multiple statements in an array but I don't think it is a massive improvement.

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
bill-warner: Agreed. Going to change the structure here

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
bill-warner: Fair point. Redesigning the functions

mobile/v1/snowflake/sql-runner/playbooks/custom/04-session-goals/01-session-goals-main.yml.tmpl
bill-warner: Good spot.

mobile/v1/snowflake/sql-runner/sql/standard/01-base/01-main/00-setup-base.sql
bill-warner: That is a fair point but I think it is slightly confusing to have this strict definition and then happily mutate it later on. I am going to remove.

mobile/v1/snowflake/sql-runner/playbooks/standard/00-setup/99-metadata-complete.yml.tmpl
bill-warner: I think this is required for the complete step.

mobile/v1/snowflake/sql-runner/playbooks/custom/04-session-goals/99-session-goals-complete.yml.tmpl
bill-warner: Good spot. Both can be

mobile/v1/snowflake/README.md
bill-warner: Nice good spot

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
bill-warner: Ok it is actually the `column_check` procedure that was changed between the web and mobile. I am removing this procedure entirely from this version of the model so I don't think there will be any naming clashes now

mobile/v1/snowflake/sql-runner/playbooks/standard/00-setup/99-metadata-complete.yml.tmpl
adatzer: The [metadata-cleanup](https://github.com/snowplow/data-models/blob/release/mobile/snowflake/1.0.0/mobile/v1/snowflake/sql-runner/sql/standard/00-setup/99-complete/99-metadata-cleanup.sql) only drops a `.scratch_schema` table, that's why the `output` one is not needed for this playbook. (btw i noticed this is something that can be removed from the respective playbook of the snowflake web model as well)

mobile/v1/snowflake/sql-runner/sql/custom/04-session-goals/99-session-goals-cleanup.sql
bill-warner: Good question. My rational here was that `session_goals_staged` is a drop and recompute table anyway, so dropping seemed a bit cleaner than truncating. The `XX-destroy` step would just be a repeat of this clean up step, hence it was removed.

bill-warner:
Ok I think I have made all the agreed upon changes + refactored the stored procedures. Interested to get your thoughts

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -46,121 +46,174 @@ CREATE OR REPLACE PROCEDURE {{.output_schema}}.mk_transaction(DML_STATEMENTS VAR
...
+                                                           SOURCE_TABLE  VARCHAR,
+                                                           TARGET_SCHEMA VARCHAR,
+                                                           TARGET_TABLE  VARCHAR)
+  RETURNS VARCHAR
+  LANGUAGE JAVASCRIPT
+  AS
+  $$
+
+  var additional_cols_stmt = `
+  WITH target_columns AS (
```
colmsnowplow: Is this roughly the same query as we use in column_checker? If so, can we possibly just call column checker here, and aggregate its results into the string we need instead?

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -46,121 +46,174 @@ CREATE OR REPLACE PROCEDURE {{.output_schema}}.mk_transaction(DML_STATEMENTS VAR
...
+  Will only alter the target table if the missing columns occur at the end of the source table.
+  Inputs:
+  SOURCE_SCHEMA:         the schema of the source table
+  SOURCE_TABLE:          the source table to select data from
+  TARGET_SCHEMA:         the schema of the target table
+  TARGET_TABLE:          the target table to delete/insert into
+  To drop:
+  DROP PROCEDURE {{.output_schema}}.alter_table(VARCHAR,VARCHAR,VARCHAR,VARCHAR);
+*/
+CREATE OR REPLACE PROCEDURE {{.output_schema}}.alter_table(SOURCE_SCHEMA VARCHAR,
```
colmsnowplow: I like the idea of separating this out into its own atomic unit. Good design choice.

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
colmsnowplow: @bill-warner `commit_table` is still best off need being renamed as it is still divergent from the function of the same name in the web models.

colmsnowplow (approved):
@bill-warner Looking really nice now - I really like what you've done refactoring the procedures, it's much cleaner now.

I do think we still need to rename `commit_table`, but aside from that, it's releasable in my book.

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
bill-warner: Yeh I looked into that but procedures in Snowflake typically don't return a result set, rather a single value. UDFs allow for result sets but I think we would have to call the UDF twice, which I think would result in two queries

bill-warner:
> @bill-warner Looking really nice now - I really like what you've done refactoring the procedures, it's much cleaner now.
> 
> I do think we still need to rename `commit_table`, but aside from that, it's releasable in my book.

Thanks. I still don't understand why we would need to rename `commit_table`. This function is not used within the web version of the model, it calls a function `commit_staged` to upsert the staged data.

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
colmsnowplow: OK - it doesn't need to be changed for this release, but this is a single value:

`{"first_result": [], "second_result": []}`

colmsnowplow:
> Thanks. I still don't understand why we would need to rename commit_table. This function is not used within the web version of the model, it calls a function commit_staged to upsert the staged data.

Ah, I was mistaken, I see what you mean now. You're correct. `commit_table` doesn't exist anywhere else, so yes it's fine to define it here, we don't require that change.

mobile/v1/snowflake/sql-runner/sql/standard/01-base/01-main/00-setup-base.sql
bill-warner: Sorry I missed this and it is a really good point. Does this mean some users atomic events table will have the column definition `se_label VARCHAR(4096)` already?

adatzer: Yes, it is so. There exist users already with `se_label` as `VARCHAR(4096)`.

bill-warner:
For clarity:
- I have removed the `mk_transaction()` procedure over security concerns and replaced with `update_manifest()` procedure.
- Refactored `alter_table()` to remove the need to query information schema for a second time. 
- Changed `session_id` in `base_session_id_manifest` table to `VARCHAR(128)` rather than `VARCHAR(36)`

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
```diff
@@ -0,0 +1,307 @@
...
+  var res = snowflake.createStatement({sqlText: column_check_stmt,
+                                       binds: [TARGET_SCHEMA, TARGET_TABLE, SOURCE_SCHEMA, SOURCE_TABLE]}
+                                      ).execute();
+  res.next();
+
+  missing_in_source = res.getColumnValue(1);
+  missing_in_target = res.getColumnValue(2);
+  columns_to_add = res.getColumnValue(3);
+
+  var column_results = [];
```
colmsnowplow: Nit: No reason to initialise and populate the var separately here.

`var column_results = [missing_in_source, missing_in_target, columns_to_add];`

In fact, I'm not sure you need a var at all (unless that's something to do with Snowflake UDFs requiring it). Can you just do this?

`return [missing_in_source, missing_in_target, columns_to_add];`

colmsnowplow (approved):
Nice. Cleaner now, and concerns about safety addressed. :)

mobile/v1/snowflake/sql-runner/sql/standard/00-setup/01-main/01-stored-procedures.sql
bill-warner: Will try this now

PR #94, opened 2021-05-11
Release/mobile/redshift/1.1.0

bill-warner:


mobile/v1/redshift/sql-runner/sql/standard/01-base/01-main/00-setup-base.sql
```diff
@@ -110,7 +110,7 @@ INSERT INTO {{.output_schema}}.{{.model}}_base_event_id_manifest{{.entropy}} (
 );
 
 CREATE TABLE IF NOT EXISTS {{.output_schema}}.{{.model}}_base_session_id_manifest{{.entropy}} (
-  session_id VARCHAR(36),
+  session_id VARCHAR(128),
```
colmsnowplow, line 113: If we change this here, should we not also change domain_sessionid and `session_id` to `VARCHAR(128)` in the definition of `events_staged`?

While we're at it - is `session_id` in that table actually the mobile `session_id`? If so should we rename it?

colmsnowplow:
Just the one comment, otherwise all looks good.

Two things to add:
1. I haven't run the tests so let's be sure we've done a thorough test loop before merge.
2. Let's make sure to highlight in the release notes and post that this is a breaking change and requires a destroy & rebuild.

mobile/v1/redshift/sql-runner/sql/standard/01-base/01-main/00-setup-base.sql
bill-warner: >If we change this here, should we not also change domain_sessionid and session_id to VARCHAR(128) in the definition of events_staged?

Sorry I was under the impression that the mobile session schema limited the session_id to 36 chars but just double checked and that isn't true. Will change to VARCHAR(128)

>While we're at it - is session_id in that table actually the mobile session_id? If so should we rename it?

Sorry I don't quite follow. Rename what exactly?

PR #98, opened 2021-05-21
Feature/web/bigquery/integration tests

bill-warner:
Includes:
* Great Expectations `integration_test` suite
* `integration_test.sh` script which runs the web model 5 times over the test dataset, in 1 day increments. During each run, it checks the outputted `events_staged` table vs the expected.
*  Once complete, it runs the GE test suite to verify the outputted derived tables vs the expected derived tables.
* The expected tables were generated using BQ web model with [this](https://github.com/snowplow/data-models/commit/d4fe03b5464ba1226138cc5bde6dfcc847117c30) commit included. This should make the output deterministic.
* The tests at present will fail due to the inclusion of the 'stray page pings' edge case, which the model currently doesn't handle deterministically. I am undecided whether to include this or not. It seems like something we should be aiming to solve and therefore should be testing for but equally having the tests fail continuously due to this isnt ideal.

### DIFF
--- a/.scripts/integration_test.sh
+++ b/.scripts/integration_test.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+
+# Expected input:
+# -b (binary) path to sql-runner binary
+# -d (database) target database for expectations
+# -a (auth) optional credentials for database target
+# -m (model) target model to run i.e. web or mobile
+
+while getopts 'b:d:a:m:' opt
+do
+  case $opt in
+    b) SQL_RUNNER_PATH=$OPTARG ;;
+    d) DATABASE=$OPTARG ;;
+    a) CREDENTIALS=$OPTARG ;;
+		m) MODEL=$OPTARG ;;
+  esac
+done
+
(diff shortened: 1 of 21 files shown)


.scripts/integration_test.sh
```diff
@@ -0,0 +1,54 @@
...
+
+  bash $script_path/run_test.sh -m $MODEL -d $DATABASE -c events_staged_integration_test_${i} || exit 1;
+
+  bash .scripts/run_config.sh -b sql-runner -c $config_dir/post_test.json -t $script_path/templates/$DATABASE.yml.tmpl -v .test/integration_tests/$MODEL/v1/${DATABASE}_variables.yml.tmpl || exit;
+
+  echo "integration_check: run $i done";
+
+done || exit 1
+
+echo "integration_check: Checking actual vs. expected for derived tables";
```
colmsnowplow, line 46: Checking the assumption - the intent is for this to run once, after all runs of the model are complete yeah?

colmsnowplow (approved):
LGTM!

.scripts/integration_test.sh
bill-warner: Yeh exactly

Issue #109, opened 2021-08-02
Snowflake web: Fix se_label column length in events_staged


adatzer:
At the moment, the Snowflake web model sets `se_label` column length to 1000: https://github.com/snowplow/data-models/blob/9bf46f3d08d0305ee23ac8146dfdf3468c5619ec/web/v1/snowflake/sql-runner/sql/standard/01-base/01-main/00-setup-base.sql#L176 (related to https://github.com/snowplow-incubator/snowplow-snowflake-loader/issues/134).

This prevents the commit step of base module to succeed when the `se_label` column is of 4096 length in `atomic.events`, which is now normally the case.

colmsnowplow:
Note that even if we're dealing with an `atomic.events` table which has it set to 1000, 4096 in the model won't break.

Issue #111, opened 2021-08-28
scratch.events_staged definition is static, scratch.events_this_run mutates with atomic.events


aidanradford:
scratch.events_staged has a written out DDL statemenhttps://github.com/snowplow/data-models/blob/master/web/v1/snowflake/sql-runner/sql/standard/01-base/01-main/00-setup-base.sql#L106

scratch.events_this_run is a table that can mutate along with the structure of atomic.events every time the procedure is run:
https://github.com/snowplow/data-models/blob/master/web/v1/snowflake/sql-runner/sql/standard/01-base/01-main/06-events-this-run.sql#L18

This mismatch between events_staged and events_this_run will lead to failure of the derived.commit_staged procedure:
https://github.com/snowplow/data-models/blob/master/web/v1/snowflake/sql-runner/sql/standard/01-base/01-main/09-commit-base.sql#L21
If:
- a user has enabled any automatic enrichements (any UA parser)
- a user has enabled performance timing
- a user has enabled the spiders and robots enrichment
- a user is tracking custom events or entities

Proposed solution:

Replace https://github.com/snowplow/data-models/blob/master/web/v1/snowflake/sql-runner/sql/standard/01-base/01-main/00-setup-base.sql#L106-L268

With:

```
-- Create staged table

CREATE OR REPLACE PROCEDURE {{.scratch_schema}}.create_events_staged()
  RETURNS VARCHAR
  LANGUAGE JAVASCRIPT
  AS
  $$

  var sql_stmt = `
      SELECT listagg(isc.column_name, ',') WITHIN GROUP (order by isc.ordinal_position)
      FROM information_schema.columns AS isc
      WHERE table_schema=UPPER('{{.input_schema}}')
        AND table_name=UPPER('events')
        AND column_name != UPPER('contexts_com_snowplowanalytics_snowplow_web_page_1');`;

  var res = snowflake.createStatement({sqlText: sql_stmt}).execute();
  res.next();
  var result = res.getColumnValue(1);

  var new_col = 'contexts_com_snowplowanalytics_snowplow_web_page_1[0]:id::varchar(36) AS page_view_id';
  if (result !== '') {
      new_col = new_col + ',';
  }

  var fin_query=`
    CREATE OR REPLACE TABLE {{.scratch_schema}}.events_staged{{.entropy}}
    AS
      SELECT
        ` + new_col + ` ` + result + `
      FROM {{.input_schema}}.events AS a
      WHERE 1=0 ;`;

[...]

PR #119, opened 2022-01-28
replace-events-staged-ddl-logic

aidanradford:
Replace scratch.events_staged table DDL logic from CREATE TABLE statement to  a stored procedure.

This will help to fix the se_label column size issue described here: https://github.com/snowplow/data-models/issues/109

and is consistent with the stored procedure used to create the events_this_run table: https://github.com/snowplow/data-models/blob/master/web/v1/snowflake/sql-runner/sql/standard/01-base/01-main/06-events-this-run.sql#L18

### DIFF
--- a/web/v1/snowflake/sql-runner/sql/standard/01-base/01-main/00-setup-base.sql
+++ b/web/v1/snowflake/sql-runner/sql/standard/01-base/01-main/00-setup-base.sql
@@ -103,166 +103,42 @@ AS (
 
 
 -- Create staged table
-CREATE TABLE IF NOT EXISTS {{.scratch_schema}}.events_staged{{.entropy}} (
 
-  page_view_id                VARCHAR(36),
+CREATE OR REPLACE PROCEDURE {{.scratch_schema}}.create_events_staged()
+  RETURNS VARCHAR
+  LANGUAGE JAVASCRIPT
+  AS
+  $$
 
-  app_id                      VARCHAR(255),
-  platform                    VARCHAR(255),
+  var sql_stmt = `
+      SELECT listagg(isc.column_name, ',') WITHIN GROUP (order by isc.ordinal_position)
+      FROM information_schema.columns AS isc
+      WHERE table_schema=UPPER('{{.input_schema}}')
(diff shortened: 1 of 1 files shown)


aidanradford:
Mentioning the issue here: https://github.com/snowplow/data-models/issues/111

colmsnowplow:
Sorry Aidan, I should've fed this back on the issue but I completely missed a key point on the proposed solution:

If I read this correctly, it'll replace the events_staged table on every run. But this is a problem - the staged table is semi-permanent. It needs to retain data in-between runs, because the design allows for one to run a module multiple times before we're done with the data in the staged table. 

Think of an asynchronous DAG, where we have the page views module - which should be up to date once a day, and we have a custom set of reports that we want to be up to date every hour. We want to run the base module once an hour, but we only want to _complete_ the base module (and overwrite the staged table) once a day. I think this solution breaks that model. 

Now, there's an argument that this is a complex use case that we don't need to necessarily support, but if that's the case we don't need both `this_run` and `staged` tables. I'd say that's a broader decision that I'd leave to @bill-warner to consider. I think this is actually how the dbt models work, so there's a case to rationalise.

Narrower term though, I _think_ your proposed solution is compatible if either a) the stored procedure only runs if the table doesn't already exist, or b) it changes to `CREATE IF NOT EXISTS` (or similar, can't remember what snowflake supports).

I am quite confused though as to _why_ we need this solution - if the issue is not the `se_label` issue, I don't see how we can have incompatible table definitions. The staged table is created at first with a static definition, but it's only the static columns that are always static. The `commit_table` function should add anything that doesn't already exist. Are we sure that there's another at play here? Have we pinned down what causes it?

aidanradford:
You're right the `commit_table` function is sufficient - the se_label column is the problem. Closing this PR.

colmsnowplow:
Good to hear @aidanradford ! Always the right call to raise a potential issue regardless. Thanks for flagging both issues back up. I've raised fixing the `se_label` one and it's on the backlog to get fixed as soon as we can fit it in. :)

Issue #129, opened 2022-05-30
Redshift web: Change SORTKEY encoding to RAW


mark-walle:
According to https://github.com/awslabs/amazon-redshift-utils/blob/master/src/Investigations/EarlyMaterialization.md it is possible that encoded sorkeys can cause "sortkey skew" and lead to excessive block I/O with early materialization.

In:

- web/v1/redshift/sql-runner/sql/standard/02-page-views/01-main/00-setup-page-views.sql
- web/v1/redshift/sql-runner/sql/standard/03-sessions/01-main/00-setup-sessions.sql
- web/v1/redshift/sql-runner/sql/standard/04-users/01-main/00-setup-users.sql

The tables:

- {{.output_schema}}.page_views{{.entropy}}
- {{.output_schema}}.sessions{{.entropy}}
- {{.output_schema}}.users{{.entropy}}

are created with explicit `ZSTD` encoding on the columns defined as primary sortkeys.

I believe these columns should have their encodings changed from `ZSTD` to `RAW`.

PR #130, opened 2022-05-30
web/v1 explicit SORTKEY encodings changed to RAW

mark-walle:
To remove risk of sortkey skew leading to excessive block I/O with early materialization per
https://github.com/awslabs/amazon-redshift-utils/blob/master/src/Investigations/EarlyMaterialization.md

issue: #129


colmsnowplow:
Cherry-picked into release branch, PR is here: https://github.com/snowplow/data-models/pull/132

@mark-walle thanks again for the PR! The UK team is back on Monday and I'd hope we can get a fairly speedy review once they're back. :)

Issue #131, opened 2022-06-02
Redshift Web: Fix column lengths in manifest tables


colmsnowplow:
In Redshift Web 1.2.0 we expanded column definitions for fields subject to pseudonymisation. We neglected to do so for the manifest tables, however.

I suspect that we hadn't run pseudonymisation on our test data yet, as this caused the models to fail during my test runs for another change.

PR #132, opened 2022-06-03
Release/web/redshift/1.3.0

colmsnowplow:
@agnessnowplow @emielver I think just a review from one of you would suffice here, unless you disagree.

You'll notice that I added 'migration' sql - this is on a suggestion from Egor which I think is a neat way to deal with manual migration steps instead of describing in release notes. This way, if someone in future needs to go from 1.0.0 to 1.7.0, they just need to run each migration playbook in order.

Not sure if there's a way to do something similar for the dbt side of things but it's worth considering IMO!

### DIFF
--- a/web/v1/redshift/sql-runner/sql/standard/99-migrations/1.3.0-migration/users-manifest.sql
+++ b/web/v1/redshift/sql-runner/sql/standard/99-migrations/1.3.0-migration/users-manifest.sql
@@ -1,0 +1,3 @@
+-- Extend column lengths to match their correct values
+-- Since ALTER TABLE ALTER COLUMN cannot run inside a multiple command statement, we need a file each.
+ALTER TABLE {{.output_schema}}.users_manifest{{.entropy}} ALTER COLUMN domain_userid type VARCHAR(128);
--- a/.test/great_expectations/expectations/web/v1/base_redshift.json
+++ b/.test/great_expectations/expectations/web/v1/base_redshift.json
@@ -286,7 +286,7 @@
   "meta": {
     "versions": {
       "test_suite_version": "1.1.1",
-      "redshift_model_version": "1.2.0"
+      "redshift_model_version": "1.3.0"
     },
     "great_expectations.__version__": "0.12.0"
   }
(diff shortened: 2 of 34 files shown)


colmsnowplow:
Note: I also had to fix https://github.com/snowplow/data-models/issues/131 in order to get the tests to run on our newly anonymised dataset.

agnes-kiss:
Hi Colm, Thanks for this, I'll review it now.

web/v1/redshift/sql-runner/sql/standard/99-migrations/1.3.0-migration/users-manifest.sql
```diff
@@ -0,0 +1,3 @@
+-- Extend column lengths to match their correct values
+-- Since ALTER TABLE ALTER COLUMN cannot run inside a multiple command statement, we need a file each.
+ALTER TABLE {{.output_schema}}.users_manifest{{.entropy}} ALTER COLUMN domain_userid type VARCHAR(128);
```
agnes-kiss, line 3: I think it is safer to add 'IF EXISTS' for the alter table statements, in my case it failed at this step as running the model with the previous version with the column length bug I did not have that table due to that failing. However, users should already have this table so it's rather nitpicking.

agnes-kiss (approved):
LGTM. Great idea on the migrations.yml, will definitely do something similar for the dbt projects.

web/v1/redshift/sql-runner/sql/standard/99-migrations/1.3.0-migration/users-manifest.sql
colmsnowplow: Can't do any harm. Done.

colmsnowplow: Actually, I tested it and I don't think `IF EXISTS` is valid for an alter table statement on Redshift (it works on snowflake). Not sure whether it's down to how I'm running it or it just not being possible, but since we don't need it I've reverted to removing `IF EXISTS` in order to get this shipped. :)

agnes-kiss: good catch, same here. Sorry about the confusion, it must have ran initially with the snowflake config.

PR #134, opened 2022-06-07
Maintenance/add migrations

colmsnowplow:
Maintenance release to follow up on the Redshift v1.3.0 release's addition of migration files, and add them for the previous releases.

One release (redshift mobile v1.1.0 https://github.com/snowplow/data-models/releases/tag/mobile%2Fredshift%2F1.1.0) specifies a complete teardown is required. I remember at the time that I believe there were major breaking changes, and this release came soon enough after the previous one that it was unlikely for people to have issues.

I don't think it's a good idea to add a playbook to tear everything down under a 'migrations' playbook, so I've left that one to be handled manually should we ever encounter it.

### DIFF
--- a/mobile/v1/snowflake/sql-runner/sql/standard/99-migrations/1.1.0-migration/mobile-events-staged.sql
+++ b/mobile/v1/snowflake/sql-runner/sql/standard/99-migrations/1.1.0-migration/mobile-events-staged.sql
@@ -1,0 +1,42 @@
+-- Create Procedure to drop mobile_events_staged if it is empty
+CREATE OR REPLACE PROCEDURE {{.scratch_schema}}.drop_mobile_events_staged_if_empty()
+  RETURNS VARCHAR
+  LANGUAGE JAVASCRIPT
+  AS
+  $$
+    var getRowCount = snowflake.createStatement(
+           {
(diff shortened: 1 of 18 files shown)


mobile/v1/snowflake/sql-runner/sql/standard/99-migrations/1.1.0-migration/mobile-events-staged.sql
```diff
@@ -0,0 +1,42 @@
...
+           }
+        );
+	
+    var res = getRowCount.execute();
+    res.next();
+    row_count = res.getColumnValue(1);
+    if (row_count > 0) {
+        throw "Ensure that all data has been processed from events_staged and it is empty.";
+    }
+    else {
```
agnes-kiss, line 19: I don't think this is dropping the table here.

colmsnowplow: 🤦 Good spot. Fixed.

agnes-kiss (approved):
LGTM! (apart from the maintenance release date in the changelog, sorry again for the delay!) As discussed, food for thought for later/dbt version to add additional comments somewhere in case variables change or there are other good to knows about the migration that does not necessarily require sql changes.

PR #137, opened 2022-09-01
Redshift Web: load_tstamp missing from table definition (Close #135)

### LINKED ISSUE #135: Redshift Web: load_tstamp missing from table definition
From [RDB Loader v4.0.0](https://github.com/snowplow/snowplow-rdb-loader/blob/master/CHANGELOG#L71) onwards, `load_tstamp` is added to the events table. This breaks the Redshift models at the insert step, where the `events_staged` table gets populated:

```sql
    INSERT INTO {{.scratch_schema}}.events_staged{{.entropy}} (
    SELECT * FROM {{.scratch_schema}}.events_this_run{{.entropy}}
```

The new column should be added to the table definitions for `events_staged` to avoid the *INSERT has more expressions than target columns* issue.

agnes-kiss:
The simplest fix for the missing load_tstamp issue. I could not test the migrations sql as I got: `ERROR: dial tcp :0: connect: can't assign requested address` and not sure what it is caused by. The model ran fine after I manually dropped the staging table first. 

Please note that the hotfix to add the field will not work as that is not the last column (but the page_view_id) so it will try to import wrong columns and error out due to the differing data types)

Also this will assume that any new customers or upgraders will be using RDB Loader >=4.0.0, perhaps it should be stated somewhere as a prerequisite or is this approach too risky overall?


colmsnowplow:
I know we did discuss this in slack, however

> Also this will assume that any new customers or upgraders will be using RDB Loader >=4.0.0, perhaps it should be stated somewhere as a prerequisite or is this approach too risky overall?

I didn't think of this part. I don't think we can safely release this patch - and I think this aspect means that this approach is not the simplest solution, in fact it complicates things a lot, because it means we cannot support bugfixes to models unless people are on a specific loader version.

I think it's better to use the `SELECT` statement in creating the table - it leads to less complication in rollout

agnes-kiss:
Agreed and thank you for your help debugging the GE section! I have run the e2e and finished just fine.

colmsnowplow (approved):
LGTM!

Don't forget to bump versions in a prepare for release commit. Good stuff getting through the various blockers on this Agnes.

PR #141, opened 2022-10-12
Snowflake Web: Optimise partition pruning (Close #140)

### LINKED ISSUE #140: Snowflake Web: Optimise partition pruning
Two of the scratch tables that select from the `events` table use a subquery to filter the data on the query limits. This may cause issues for partition pruning in Snowflake as described in [this](https://community.snowflake.com/s/article/Pruning-is-not-happening-subquery) article. 

Affected tables:

```
{{.scratch_schema}}.base_sessions_to_process{{.entropy}}
{{.scratch_schema}}.events_this_run{{.entropy}}
```

To improve partition pruning the subquery needs to be replaced with alternative SQL methods (e.g.: by abstracting them to variables during a stored procedure)

agnes-kiss:
As the title says, partition pruning needs to be improved within the Snowflake web model. Local tests showed significant improvement.

PR check with Great Expectations ran fine.

### DIFF
--- a/web/v1/snowflake/sql-runner/sql/standard/01-base/01-main/03-sessions-to-process.sql
+++ b/web/v1/snowflake/sql-runner/sql/standard/01-base/01-main/03-sessions-to-process.sql
@@ -15,6 +15,10 @@
 */
 
 
+-- Use variable to set scan limits
+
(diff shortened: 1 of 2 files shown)


emielver (approved):
LGTM! Great job on the turnaround time on this!

emielver:
Oh one small thing, should we do the same for the mobile model just because?

agnes-kiss:
Yeah, we might as well just do it now too. Will go ahead and prepare that one so that I can release them one after the other.

PR #143, opened 2022-12-13
Release/web/bigquery/1.0.4

### LINKED ISSUE #75: BigQuery Web: Fix upsert limit in commit_table procedure
Upsert limits calculated [here](https://github.com/snowplow/data-models/blob/e1e10fb696a52aa9bc990e975154405779f421de/web/v1/bigquery/sql-runner/sql/standard/00-setup/01-main/01-functions-and-procedures.sql#L108) should use the `SOURCE_PATH` var not `TARGET_PATH`.

agnes-kiss:
Patch release for: BigQuery Web: Fix upsert limit in commit_table procedure (Close #75)

This is to make sure the **upsert_lookback_days** variable works as it was intended to.

```
upsert_lookback_days:  default 30. Period of time (in days) to look back over the production table in order to find rows to delete when upserting data. Where performance is not a concern, should be set to as long a value as possible.
```
PR check with Great Expectations ran fine.

### DIFF
--- a/.test/great_expectations/expectations/web/v1/base.json
+++ b/.test/great_expectations/expectations/web/v1/base.json
@@ -138,7 +138,7 @@
   "meta": {
     "versions": {
       "test_suite_version": "1.1.1",
-      "bigquery_model_version": "1.0.3",
+      "bigquery_model_version": "1.0.4",
       "snowflake_model_version": "1.0.2"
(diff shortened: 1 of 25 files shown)


colmsnowplow (approved):
LGTM! My bad for the original bug!

